PR #19, opened 2019-10-22
Add isLayerReady check in LocationComponent before updating compass listener.

pengdev:
Resolves https://github.com/mapbox/mapbox-gl-native/issues/14889

### DIFF
--- a/MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationComponentTest.kt
+++ b/MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationComponentTest.kt
@@ -387,6 +387,23 @@ class LocationComponentTest {
   }
 
   @Test
+  fun compass_notAddListenerWhenLayerNotReady() {
+    locationComponent.activateLocationComponent(context, mock(Style::class.java), locationEngine, locationEngineRequest, locationComponentOptions)
+    locationComponent.onStart()
+    locationComponent.isLocationComponentEnabled = true
+    `when`(mapboxMap.cameraPosition).thenReturn(CameraPosition.DEFAULT)
+    `when`(locationLayerController.isConsumingCompass).thenReturn(true)
+    locationComponent.renderMode = RenderMode.COMPASS
+
+    verify(compassEngine, times(1)).addCompassListener(any(CompassListener::class.java))
+
+    locationComponent.onStartLoadingMap()
+    // Layer should be disabled at this point
+    locationComponent.setCameraMode(CameraMode.TRACKING_COMPASS)
+    verify(compassEngine, times(1)).addCompassListener(any(CompassListener::class.java))
+  }
+
+  @Test
   fun developerAnimationCalled() {
     locationComponent.activateLocationComponent(context, mock(Style::class.java), locationEngine, locationEngineRequest, locationComponentOptions)
     locationComponent.isLocationComponentEnabled = true
--- a/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/location/LocationModesActivity.java
+++ b/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/location/LocationModesActivity.java
@@ -233,8 +233,10 @@ public class LocationModesActivity extends AppCompatActivity implements OnMapRea
       return;
     }
 
-    String styleUrl = Style.DARK.equals(mapboxMap.getStyle().getUri()) ? Style.LIGHT : Style.DARK;
-    mapboxMap.setStyle(new Style.Builder().fromUri(styleUrl));
+    mapboxMap.getStyle(style -> {
+      String styleUrl = Style.DARK.equals(style.getUri()) ? Style.LIGHT : Style.DARK;
+      mapboxMap.setStyle(new Style.Builder().fromUri(styleUrl));
+    });
   }
 
   private void disableGesturesManagement() {
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
@@ -1289,7 +1289,7 @@ public final class LocationComponent {
         return;
       }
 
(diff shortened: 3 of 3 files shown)


MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/location/LocationModesActivity.java
```diff
@@ -233,8 +233,10 @@ private void toggleMapStyle() {
       return;
     }
 
-    String styleUrl = Style.DARK.equals(mapboxMap.getStyle().getUri()) ? Style.LIGHT : Style.DARK;
-    mapboxMap.setStyle(new Style.Builder().fromUri(styleUrl));
+    mapboxMap.getStyle(style -> {
```
LukasPaczos: NIT: testing leftover?

pengdev:
While trying to reproduce the user reported issue, I have found another issue related to trying to use the style while it's in invalidate state. Created another issue to follow up at https://github.com/mapbox/mapbox-gl-native-android/issues/20

MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationComponentTest.kt
```diff
@@ -386,6 +386,24 @@ class LocationComponentTest {
...
+  @Test
+  fun compass_notAddListenerWhenLayerNotReady() {
+    locationComponent.activateLocationComponent(context, mock(Style::class.java), locationEngine, locationEngineRequest, locationComponentOptions)
+    locationComponent.isLocationComponentEnabled = true
+    locationComponent.onStart()
+    `when`(locationLayerController.isConsumingCompass).thenReturn(true)
+
+    locationComponent.onStartLoadingMap()
+    // Layer should be disabled at this point
+    locationComponent.setCameraMode(CameraMode.TRACKING)
```
LukasPaczos: NIT: It'd be better to use a mode that actually consumes the compass, like `setRenderMode(RenderMode.COMPASS)`. It's not impacting the test because of the `locationLayerController` mock above but it would be good to correct.

MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationComponentTest.kt
```diff
@@ -386,6 +386,24 @@ class LocationComponentTest {
...
+    verify(compassEngine, never()).addCompassListener(any(CompassListener::class.java))
+
+    locationComponent.onFinishLoadingStyle()
+    // Layer should be enabled at this point
+    locationComponent.setCameraMode(CameraMode.TRACKING_GPS)
```
LukasPaczos: This shouldn't be necessary, the listener should be added automatically when the style is loaded based on the previous data.

MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/location/LocationModesActivity.java
pengdev: It's a general improvement to the test app. It's a better practice to use async `getStyle() ` callback.

LukasPaczos: Ah, the `style` is for `getUri`. Didn't spot at first :+1:

MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationComponentTest.kt
pengdev: Yes, this additional verify is added to cover the case where the `isLayerReady` is changed from false to true, the addListener should go through.

MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationComponentTest.kt
pengdev: Adjusted accordingly.

LukasPaczos: Still show up the same for me :upside_down_face: Did you push all of the local changes?

pengdev: Sorry I misunderstood your first comment. I was trying to reproduce the issue you mentioned in https://github.com/mapbox/mapbox-gl-native/issues/14889#issuecomment-541128492 , where the compass listeners are re-enabled when the camera is changed before the style is loaded. Firstly I call `locationComponent.onStartLoadingMap()`, it will change the `isLayerReady` flag to false, then I change the camera mode, internally, it would call `updateCompassListenerState()`, and I will verify that the compass listeners are not added because the Layer is not ready.

LukasPaczos: All is fine, just that `CameraMode.TRACKING` doesn't actually use compass data. That's taken into account in the camera implementation itself, but if it were to change, this test might not hit the target anymore. Anyway, we can address later if needed.

pengdev: sorry @LukasPaczos I just got it, you meant change to something like `locationComponent.setCameraMode(CameraMode.TRACKING_COMPASS)` not `setRenderMode(RenderMode.COMPASS)`, will refactor to it.

oantajames:
@LukasPaczos when will this be released? having some random crashes...

LukasPaczos:
@oantajames you can test the changes with `v8.5.0-beta.1`. The stable release should go out later this week.

oantajames:
any updates on the stable release? @LukasPaczos

tobrun:
@oantajames it has been released, see https://search.maven.org/artifact/com.mapbox.mapboxsdk/mapbox-android-sdk/8.5.0-beta.1

oantajames:
Thanks @tobrun

PR #24, opened 2019-10-24
Update mapbox-gl-native submodule.

tarigo:


tarigo:
> we just had a couple more merges, can we update to 646d4e3996513ed3e9247e958d39ce9e9b506652 ?

Yep.

Issue #31, opened 2019-10-28
LocalGlyphRasterizer and high contrast accessibility 


tobrun:
When you have map supporting CJK characters, generated with local glyph rasterizer, and you enable the high contrast accessibility setting. Characters become unreadable:

![image](https://user-images.githubusercontent.com/2151639/67659760-70210e00-f95d-11e9-8136-1a6c6423c774.png)

More info of high contrast [here](https://support.google.com/accessibility/android/answer/6151855?hl=en)

tobrun:
We can detect if accessibility manager has this setting enabled with:

```
AccessibilityManager am = (AccessibilityManager) this.getSystemService(Context.ACCESSIBILITY_SERVICE);
boolean isHighTextContrastEnabled = am.isHighTextContrastEnabled();
```

ystsoi:
I found that this issue can be worked around by drawing text over white background, and then get the RGB values instead of the alpha values.

That is, in [LocalGlyphRasterizer.java](https://github.com/mapbox/mapbox-gl-native-android/blob/master/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java), change:
```
canvas.drawColor(Color.TRANSPARENT, PorterDuff.Mode.CLEAR);
```
to:
```
canvas.drawColor(Color.WHITE);
```

In [local_glyph_rasterizer.cpp](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/src/text/local_glyph_rasterizer.cpp), change:
```
fixedMetrics.bitmap.data[i] = rgbaBitmap.data[4 * i + 3];
```
to:
```
fixedMetrics.bitmap.data[i] = 0xff - round(0.2126 * rgbaBitmap.data[4 * i] + 0.7152 * rgbaBitmap.data[4 * i + 1] + 0.0722 * rgbaBitmap.data[4 * i + 2]);
```

In most cases, RGB components have the same value. Converted the RGB values into grayscale values, just in case there are color symbols.

The result should be the same as the previous logic in most cases.

tobrun:
@ystsoi thank you for the thorough analysis and providing feedback!
I will try finding some time testing this out!

cc @Chaoba

PR #50, opened 2019-11-11
Disable local font generation with accessibility high contrast mode

tobrun:
Refs #31, mitigates the issue from #32 by disabling local glyph generation when this accessibility feature is active. In this setup, it will download the glyphs online instead of generating them corrupt on device.


tobrun:
obsolete with https://github.com/mapbox/mapbox-gl-native-android/pull/289

PR #64, opened 2019-11-22
[offline] expose packDatabase

tobrun:
This updates to latest of gl-native master and brings in changes for https://github.com/mapbox/mapbox-gl-native/pull/15899.


chloekraw:
hmm @tobrun why is doing this important now as opposed to after we add the other API that decouples ambient cache clearing from `VACUUM`?

tobrun:
No reason @chloekraw , we need to periodically update gl-native submodule and pull in all required changes from time to time. I was already seeing 90 commits since last update. Expect such PRs to land on a weekly basis. The longer we wait, the higher chance we see breaking changes (changes that require to align the code here vs code in gl-native).

tobrun:
>  90 commits since last update
>> +1 to bump. Just a note that packDatabase has already been brought in a couple of days ago in dfdbaf5.

I think, I might have had an stale master :sweat_smile:

PR #65, opened 2019-11-22
Expose MapView configuration to not activate/deactivate

tobrun:
This PR is solely created for testing purposes.


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -401,7 +417,7 @@ public void onStop() {
       mapRenderer.onStop();
     }
 
-    if (isStarted) {
+    if (isStarted && deactivateFileSource) {
```
LukasPaczos: If we opt to no deactivate FileSource and ConnectivityReceiver here, we should do that in `onDestroy`.

tobrun:
This is no longer an issue for a specific customer. closing

PR #80, opened 2019-12-03
Invoke location camera transition finish callback when equal mode is set

LukasPaczos:
Ensure that we notify `OnLocationCameraTransitionListener`s when the same mode is reset.


LukasPaczos:
Changelog in https://github.com/mapbox/mapbox-gl-native-android/pull/85.

PR #83, opened 2019-12-03
Invoke cancelable callback when zoom/tilt while tracking is ignored

LukasPaczos:
To make the callbacks more reliable, we should ensure to invoke them even when the transition is ignored.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
@@ -739,14 +739,15 @@ public final class LocationComponent {
                                 @Nullable MapboxMap.CancelableCallback callback) {
     checkActivationState();
     if (!isLayerReady) {
+      notifyUnsuccessfulCameraOperation(callback, null);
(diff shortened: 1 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -739,14 +739,15 @@ public void zoomWhileTracking(double zoomLevel, long animationDuration,
                                 @Nullable MapboxMap.CancelableCallback callback) {
     checkActivationState();
     if (!isLayerReady) {
+      notifyUnsuccessfulCameraOperation(callback, null);
```
tobrun: Would it be useful to add a message for this occurrence as well? or would that spam the user?

LukasPaczos: I'd rather skip it as this would happen only when component is paused, so it's expected that it'll ignored the updates.

LukasPaczos:
Chagelog in https://github.com/mapbox/mapbox-gl-native-android/pull/85.

Issue #82, opened 2019-12-03
Use synchronous OkHttp


tobrun:
Noticing that we use asynchronous OkHttp setup while the HttpRequesImpl will get executed on a worker thread. Since there is no need for async setup we can rework our setup to be synchronous. We can run benchmarks to validate if we get any performance improvements from these. 

cc @mapbox/maps-android

maxtracking:
@tobrun Any updates on this? Also, maybe it's time to update OkHttp version to their latest as well?

tobrun:
It's expected behavior from gl-native that a http request will spawn it's own thread. If we use the gl-native provided thread, we can't request multiple resources at once. The filsource will wait to schedule a new request until the previous request has completed.

> `update OkHttp version`

@maxtracking sadly we can't, current OkHttp version is a LTS version that works on API 14+ 
We can't bump it as our min version would become lolipop.

zugaldia:
While we can't bump the OkHttp version bundled with the SDK, the developer can always invoke [`HttpRequestUtil.setOkHttpClient(OkHttpClient client)`](https://docs.mapbox.com/android/api/map-sdk/9.0.0/com/mapbox/mapboxsdk/module/http/HttpRequestUtil.html#setOkHttpClient-OkHttpClient-) to set their own client/version.

PR #88, opened 2019-12-03
Refactor HttpRequest to execute OkHttp synchronously

tobrun:
Closes https://github.com/mapbox/mapbox-gl-native-android/issues/82, there is no need to execute the http request asychrounosly as we are already on a worker thread. This change has potential performance improvements, after this change is merged we can validate this with out metric system. Will communicate any findings back on this PR.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/http/HttpRequestImpl.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/http/HttpRequestImpl.java
@@ -6,6 +6,7 @@ import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
 import android.text.TextUtils;
 import android.util.Log;
+
 import com.mapbox.mapboxsdk.BuildConfig;
 import com.mapbox.mapboxsdk.constants.MapboxConstants;
 import com.mapbox.mapboxsdk.http.HttpIdentifier;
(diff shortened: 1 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/http/HttpRequestImpl.java
```diff
@@ -155,27 +112,54 @@ public void onResponse(@NonNull Call call, @NonNull Response response) {
...
-    private int getFailureType(Exception e) {
-      if ((e instanceof NoRouteToHostException) || (e instanceof UnknownHostException) || (e instanceof SocketException)
-        || (e instanceof ProtocolException) || (e instanceof SSLException)) {
-        return CONNECTION_ERROR;
-      } else if ((e instanceof InterruptedIOException)) {
-        return TEMPORARY_ERROR;
-      }
-      return PERMANENT_ERROR;
+  @Override
+  public void cancelRequest() {
```
LukasPaczos: Which thread manages and cancels the requests? Is it the same thread that creates the requests? Can it lead to deadlocks or prevent actual cancellation?

tobrun: I'm observing the following in both this as old implementation. If you pan the map to a new area that requires loading tiles:

First I see:
```
[HTTP] Request was successful (code = 200). https://api.mapbox.com/v4/mapbox.mapbox-streets-v8,mapbox.mapbox-terrain-v2/5/18/15.vector.pbf?access_token=pk.eyJ1IjoidG9icnVuIiwiYSI6ImNpajVlajR0cjAwNjN2NmtyY204eWw2eG0ifQ.x3_WEoExNW5Qyv9T3Vj7Mw&sku=100k3qaidfqebac703879074193acfa72f8a283b183
```

Followed by:
```
CancelRequest for call: https://api.mapbox.com/v4/mapbox.mapbox-streets-v8,mapbox.mapbox-terrain-v2/5/18/15.vector.pbf?access_token=pk.eyJ1IjoidG9icnVuIiwiYSI6ImNpajVlajR0cjAwNjN2NmtyY204eWw2eG0ifQ.x3_WEoExNW5Qyv9T3Vj7Mw&sku=100k3qaidfqebac703879074193acfa72f8a283b183
```

I initially thought this wasn't expected behavior but we do execute cancel as part of the destructor of HTTPRequest. When I optimise the code to only excute cancel when we haven't received a response. I'm seeing that the old implementation does occasionally cancel a request while this PR doesn't.

tobrun: I will follow up with a patch for mapbox-gl-native (for http_request.ccp) but will close this PR as not actionable.

PR #71, opened 2019-11-27
Add SoLoader's proguard rule

LukasPaczos:
Closes https://github.com/mapbox/mapbox-gl-native-android/issues/60.

PR #89, opened 2019-12-04
Update changelog for beta.1 for release-tequila.

pengdev:
This PR updates the changelog 8.6.0-alpha.3


pengdev:
> #71 would be great to include as well, the note can be:
> 
> ```
> Fixed a runtime crash of Proguard optimized builds on < API 21 devices by including the missing SoLoader flag
> ```

Good call 👍 , added.

Issue #81, opened 2019-12-03
Reusable visibleregion API


tobrun:
To avoid memory churn in calling the `getVisibleRegion` API many times as part of a map change event. This API should allow passing an object so it can be reused vs instantiating a new object every API call.

PR #109, opened 2019-12-17
Add getVisibleCoordinateBounds method.

pengdev:
Resolves https://github.com/mapbox/mapbox-gl-native-android/issues/81 , refs https://github.com/mapbox/mapbox-gl-native/pull/16069

This PR add `void getVisibleCoordinateBounds(@NonNull double[] bounds)` method to get visible screen coordinate bounds in an array.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLngBounds.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLngBounds.java
@@ -372,9 +372,8 @@ public class LatLngBounds implements Parcelable {
   }
 
   private boolean containsLongitude(final double longitude) {
-    return ((longitude <= this.longitudeEast) && (longitude >= this.longitudeWest))
-            || ((longitude <= this.longitudeEast - 360.0) && (longitude >= this.longitudeWest - 360.0))
-            || ((longitude <= this.longitudeEast + 360.0) && (longitude >= this.longitudeWest + 360.0));
+    return (longitude <= this.longitudeEast)
+            && (longitude >= this.longitudeWest);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLngBounds.java
```diff
@@ -372,8 +372,9 @@ private boolean containsLatitude(final double latitude) {
   }
 
   private boolean containsLongitude(final double longitude) {
-    return (longitude <= this.longitudeEast)
-      && (longitude >= this.longitudeWest);
+    return ((longitude <= this.longitudeEast) && (longitude >= this.longitudeWest))
+            || ((longitude <= this.longitudeEast - 360.0) && (longitude >= this.longitudeWest - 360.0))
+            || ((longitude <= this.longitudeEast + 360.0) && (longitude >= this.longitudeWest + 360.0));
   }
```
pengdev: @tobrun I'm not totally sure about this modification, do you see any issues with it? since previously this method doesn't consider the case of the unwrapped bounds, will this cause unexpected behaviour?

LukasPaczos: The strategy for this class was that it doesn't handle bounds wrapping, so this change might actually be unexpected.

pengdev: @LukasPaczos  thanks for the comment, so I guess I should leave it as it is now, and update my unit tests to unwrap the LatLngCoordinates before using the the contains method. 
Or is it better that I introduce a new method `containsLongitudeUnwrapped`  with the bounds wrapping support?

tobrun: I missed this change, this could indeed result in unexpected changes for users.

> update my unit tests to unwrap the LatLngCoordinates before using the the contains method.

That would be my preference for now, if there is an actual need for `containsLongitudeUnwrapped` we can revisit later.

pengdev: 👍 Reverted changes in LatLngBounds#contains.

PR #138, opened 2020-01-09
Update release-tequila branch for v8.6.1

tobrun:
Cherry-picks #137 and updates vendor/mapbox-gl-native to include latest from upstream `release-tequila`.


tobrun:
Buck CI job can be ignored, correct config was only added to master after branch cut of release-tequila.

PR #148, opened 2020-01-17
[sku] append sku token to offline download requests

tobrun:
This PR changes the logic of offline downloads to include the SKU token.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HttpRequestUrl.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HttpRequestUrl.java
@@ -6,6 +6,8 @@ import com.mapbox.mapboxsdk.Mapbox;
 
 public class HttpRequestUrl {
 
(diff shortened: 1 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HttpRequestUrl.java
```diff
@@ -22,20 +24,20 @@ private HttpRequestUrl() {
    * @return the adapted resource url
    */
   public static String buildResourceUrl(@NonNull String host, String resourceUrl, int querySize, boolean offline) {
+    requestBuilder.delete(0, requestBuilder.length());
```
pengdev: We could also use `requestBuilder.setLength(0)` to reset the StringBuilder.

tobrun: ah cool, wasn't aware that this is a better approach (ref https://stackoverflow.com/a/30190425).

PR #156, opened 2020-01-22
Expose disabling horizontall scroll gestures

tobrun:
This PR adds an option to disable horizontal scroll gestures through xml attributes or UiSettings.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
@@ -7,8 +7,10 @@ import android.content.Context;
 import android.content.res.Resources;
 import android.graphics.PointF;
 import android.os.Handler;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+
 import android.view.InputDevice;
 import android.view.MotionEvent;
 import android.view.animation.DecelerateInterpolator;
@@ -423,9 +425,6 @@ final class MapGestureDetector {
         return false;
       }
 
-      transform.cancelTransitions();
-      cameraChangeDispatcher.onCameraMoveStarted(REASON_API_GESTURE);
-
       // tilt results in a bigger translation, limiting input for #5281
       double tilt = transform.getTilt();
       double tiltFactor = 1.5 + ((tilt != 0) ? (tilt / 10) : 0);
@@ -434,6 +433,17 @@ final class MapGestureDetector {
 
       // calculate animation time based on displacement
       long animationTime = (long) (velocityXY / 7 / tiltFactor + MapboxConstants.ANIMATION_DURATION_FLING_BASE);
+      if (!uiSettings.isHorizontalScrollGesturesEnabled()) {
+        // determine if angle of fling is valid for performing a vertical fling
+        double angle = Math.abs(Math.toDegrees(Math.atan(offsetX / offsetY)));
+        if (angle > MapboxConstants.ANGLE_THRESHOLD_IGNORE_VERTICAL_FLING) {
+          return false;
+        }
+        offsetX = 0.0;
+      }
+
+      transform.cancelTransitions();
+      cameraChangeDispatcher.onCameraMoveStarted(REASON_API_GESTURE);
 
       // update transformation
       transform.moveBy(offsetX, offsetY, animationTime);
@@ -475,6 +485,11 @@ final class MapGestureDetector {
       if (distanceX != 0 || distanceY != 0) {
         // dispatching camera start event only when the movement actually occurred
         cameraChangeDispatcher.onCameraMoveStarted(CameraChangeDispatcher.REASON_API_GESTURE);
+
+        // Disable scrolling horizontal if not allowed
+        if (!uiSettings.isHorizontalScrollGesturesEnabled()) {
+          distanceX = 0;
+        }
 
         // Scroll the map
         transform.moveBy(-distanceX, -distanceY, 0 /*no duration*/);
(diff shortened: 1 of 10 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
```diff
@@ -435,6 +441,14 @@ public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float ve
       // calculate animation time based on displacement
       long animationTime = (long) (velocityXY / 7 / tiltFactor + MapboxConstants.ANIMATION_DURATION_FLING_BASE);
 
+      if (!uiSettings.isHorizontalScrollGesturesEnabled()) {
+        offsetX = 0.0;
+        if (offsetY < MapboxConstants.Y_TRANSLATION_THRESHOLD_IGNORE_FLING) {
```
tobrun: added this option to ignore flings if they happen mostly horizontally, the value is not really backed up by any scientific measurement.

tobrun: thinking about this today, I should calculate the angle of the fling against the horizontal axis and determine a threshold for that given angle to not execute a fling

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
```diff
@@ -312,7 +314,11 @@ boolean onGenericMotionEvent(MotionEvent event) {
           float scrollDist = event.getAxisValue(MotionEvent.AXIS_VSCROLL);
 
           // Scale the map by the appropriate power of two factor
-          transform.zoomBy(scrollDist, new PointF(event.getX(), event.getY()));
+          float horizontalScroll = event.getX();
+          if (!uiSettings.isHorizontalScrollGesturesEnabled()) {
+            horizontalScroll = 0;
+          }
+          transform.zoomBy(scrollDist, new PointF(horizontalScroll, event.getY()));
```
LukasPaczos: I think this can be reverted. This is only for mouse-wheel zoom and the `event.getX()` here is the focal point.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
```diff
@@ -434,6 +440,14 @@ public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float ve
 
       // calculate animation time based on displacement
       long animationTime = (long) (velocityXY / 7 / tiltFactor + MapboxConstants.ANIMATION_DURATION_FLING_BASE);
+      if (!uiSettings.isHorizontalScrollGesturesEnabled()) {
+        // determine if angle of fling is valid for performing a vertical fling
+        double angle = Math.abs(Math.toDegrees(Math.atan(offsetX / offsetY)));
+        if (angle > MapboxConstants.ANGLE_THRESHOLD_IGNORE_VERTICAL_FLING) {
```
LukasPaczos: Instead of doing this calculation, could you only check the `velocityY` (and come up with a new threshold) like above?

tobrun: That was the initial implementation, see https://github.com/mapbox/mapbox-gl-native-android/pull/156#discussion_r369763501. IMO this is a cleaner solution as we calculate where the fling is going, just checking the velocityY would result in ignore short vertical flings.

LukasPaczos: I think that's okay, we shouldn't really care how fast horizontally is the fling and currently, it will impact the calculation in 
https://github.com/mapbox/mapbox-gl-native-android/pull/156/files#diff-40fbbd818189e7d34788f0a24144c366R422-R426.

My idea would be to ignore the above check and replace it with one that checks only the `velocityY` (not `offsetY`) if the horizontal scroll is disabled.

Also, the verification in any form has to be moved up, so that we are not dispatching any events if it fails.

tobrun: > I think that's okay

I disagree, we have more certainty about how the gesture has been performed. Doesn't make sense saying with going with a less certain solution. 

> we shouldn't really care how fast horizontally is the fling 

I don't care either, I only care about the angle of the gesture

> and currently, it will impact the calculation in

Please explain how this would impact that calculation

>   Also, the verification in any form has to be moved up, so that we are not dispatching any events if it fails.

sure

tobrun: Tested out your requested changes in `tvn-alternative-approach` and the UX is worse. If you have large fling, even with not much y displacement the values of velocityY sky rocket.

LukasPaczos: Thanks @tobrun. I tested as well and the current setup feels alright :+1:

tobrun:
> One issue that remains is that scaling or using double-tap/two-tap to zoom in/out is going to impact the horizontal component of the camera position because we're using the focal point of a gesture. 

For the specific use-case that we are adding this, they would probably disable all other gestures except scroll. Proposed solution is however better, tracking this in #157. Users can work around this with setting a fixed focal point themselves.

PR #166, opened 2020-02-06
Expose option to pass intermediate points for smooth location animation

LukasPaczos:
This PR adds a `LocationComponent#forceLocation` overload, which allows passing a list of locations. The last location in the list is the desired target, while everything else makes up for the intermediate points in the otherwise linear location puck animation. This allows for better road geometry tracking and avoids "cutting" corners when 2 raw updates are on the opposite side of an intersection maneuver.

This also exposes a `lookAheadUpdate` flag in the same method. If we have an ability to predict location updates, setting this flag to `true` will take the timestamp of the target location (last in the list) and calculate the animation duration equal to
```
targetLocationTimestamp - currentTimestamp
```
This allows minimizing the visual lag that a smooth animation would introduce otherwise.

/cc @SiarheiFedartsou @mskurydin

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
@@ -3,11 +3,13 @@ package com.mapbox.mapboxsdk.location;
 import android.animation.Animator;
 import android.location.Location;
 import android.os.SystemClock;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.annotation.VisibleForTesting;
 import android.util.SparseArray;
 import android.view.animation.LinearInterpolator;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.Size;
+import androidx.annotation.VisibleForTesting;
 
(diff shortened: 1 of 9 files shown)


chloekraw:
@LukasPaczos neat idea. do we have any metrics on the performance diff with this change?

LukasPaczos:
I don't have data, but there should be no perf difference. Normally, the framework would interpolate between A->B with an arbitrary amount of intermediate points (system determines that based on the framerate). While now, the system will still fill in some intermediate points, but we're also making it possible to define a fixed amount of them which the system will have to visit giving us a chance to better shape the path.

Guardiola31337:
https://github.com/mapbox/mapbox-gl-native-android/pull/166/commits/24ed3be62696ee67fcdfc0a2734283d0f1f72efb updates `mapbox-gl-native` vendor to point to https://github.com/mapbox/mapbox-gl-native/commit/3ea459233a60ee737d0b637582f94024f3763797 (https://github.com/mapbox/mapbox-gl-native/issues/16180#issuecomment-582899890) that fixes the crash we were seeing in https://github.com/mapbox/mapbox-gl-native/issues/16180 https://github.com/mapbox/mapbox-gl-native/issues/16101 and https://github.com/mapbox/mapbox-navigation-android/pull/2452#issuecomment-583567113 

👀 @tobrun @LukasPaczos

LukasPaczos:
Thanks @Guardiola31337 :+1:

PR #171, opened 2020-02-12
Add expression in.

Chaoba:
This pr add the interface for expression `in`. Not include the corresponding [pr](https://github.com/mapbox/mapbox-gl-native/pull/16162) in core sdk now.
@mapbox/maps-android any suggestion on how to test it?

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
@@ -1537,6 +1537,42 @@ public class Expression {
   }
 
   /**
+   * Retrieves whether an item exists in an array or a substring exists in a string.
+   *
+   * @param needle     the item expression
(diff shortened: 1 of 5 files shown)


Chaoba:
> You can test your changes by updating the `vendor/mapbox-gl-native` submodule pin to your PR in mapbox-gl-native. Once that is merged we can set the submodule pin to master.

That's right. We'd better add an option in `RuntimeStyleActivity` to test express in, but I am not sure how it should be used in app. Can you give an example about it?

tobrun:
```
 ["in", needle: (boolean, string or number), haystack: (array or string)]: boolean
```

The returned boolean value can be used with case expression:

```
["case",
    condition: boolean, output: OutputType,
    condition: boolean, output: OutputType,
    ...,
    fallback: OutputType
]: OutputType
```

So you could make a textSize of a certain city bigger if it contained the word "york":
 - New york
 - York

Chaoba:
It seems works well. 
device-2020-02-13-141730

chloekraw:
cc @mapbox/maps-ios @alexshalamov

PR #198, opened 2020-02-26
Add expression within

### LINKED ISSUE #191: Platform bindings for `within` expression
Tracking the development of platform bindings for the new `within` expression (https://github.com/mapbox/mapbox-gl-native/pull/16157; https://github.com/mapbox/mapbox-gl-native/pull/16194)

If it's possible, because this isn't too complex, it would be great to get this merged this week before the next `alpha` release.

cc @mapbox/maps-android @zmiao

Chaoba:
Resolves #191 

This expression will cause error:
 ` E/Mbgl: {pboxsdk.testapp}[JNI]: Error setting property: text-size [1]: 'within' expression requires valid geojson source that contains polygon geometry type.` 

The error is triggered by `isObject` method return false  in this line https://github.com/mapbox/mapbox-gl-native/blob/9aefd2f2faf0e66afc263e96e3cff4cc03295e64/src/mbgl/style/expression/within.cpp#L111

The geojson passed into core is `{“type”:“Polygon”,“coordinates”:[[[-13.359375,67.6092206],[-14.0625,36.1733569],[43.59375,36.456636],[-13.359375,67.6092206]]]}` and @zmiao already confirmed that this geojson could pass glfw test.

@mapbox/maps-android any idea about this error?

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
@@ -12,6 +12,7 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
+import com.mapbox.geojson.GeoJson;
 import com.mapbox.mapboxsdk.style.layers.PropertyFactory;
 import com.mapbox.mapboxsdk.style.layers.PropertyValue;
 
@@ -1570,6 +1571,11 @@ public class Expression {
    */
   public static Expression in(@NonNull String needle, @NonNull Expression haystack) {
     return new Expression("in", literal(needle), haystack);
+  }
+
+  public static Expression within(@NonNull GeoJson geojson) {
+    String string = geojson.toJson();
+    return new Expression("within", literal(string));
   }
 
   /**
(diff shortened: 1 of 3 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
```diff
@@ -1563,15 +1565,24 @@ public static Expression in(@NonNull Number needle, @NonNull Expression haystack
...
+   * @param needle   the item expression
+   * @param haystack the array or string expression
    * @return true if exists.
    * @see <a href="https://www.mapbox.com/mapbox-gl-js/style-spec/#expressions-in">Style specification</a>
    */
   public static Expression in(@NonNull String needle, @NonNull Expression haystack) {
     return new Expression("in", literal(needle), haystack);
   }
 
+  public static Expression within(@NonNull Polygon polygon) {
```
tobrun: pretty cool that you got it to work with a geojson type. This is a very handy utility function. To be feature complete we will need to expose `within(@NonNull Expression expression)` so developers can use other expressions in conjunction with this one.

tobrun: update: now with looking more into upstream PR, we do need to expose only the Polygon variation. Expressions will now support inline geometry!

Chaoba:
Change country label opacity. 
device-2020-02-28-181404

PR #222, opened 2020-03-05
Provide a way to opt out from LocationComponent initialisation

maxtracking:
On the low-spec custom hardware where every tiny bit counts in resources, we have use-cases when standard location component is never used as it does not fit the requirements.

This PR introduces the following:
- In use-cases when LocationComponent is never used, provide opt-out option
- Do not inject LocationComponent when opted out

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -13,12 +13,6 @@ import android.view.View;
 import android.widget.FrameLayout;
 import android.widget.ImageView;
 
-import androidx.annotation.CallSuper;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.annotation.UiThread;
-import androidx.collection.LongSparseArray;
-
 import com.mapbox.android.gestures.AndroidGesturesManager;
 import com.mapbox.mapboxsdk.MapStrictMode;
 import com.mapbox.mapboxsdk.Mapbox;
@@ -42,6 +36,12 @@ import java.util.List;
 
 import javax.microedition.khronos.egl.EGLConfig;
 import javax.microedition.khronos.opengles.GL10;
+
+import androidx.annotation.CallSuper;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.UiThread;
+import androidx.collection.LongSparseArray;
 
 import static com.mapbox.mapboxsdk.maps.widgets.CompassView.TIME_MAP_NORTH_ANIMATION;
 import static com.mapbox.mapboxsdk.maps.widgets.CompassView.TIME_WAIT_IDLE;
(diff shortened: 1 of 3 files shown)


chloekraw:
@maxtracking thanks for the contribution! My initial reaction is that this sounds like a good idea, we'll dig in further. cc @zugaldia @LukasPaczos

LukasPaczos:
hey @maxtracking, thanks a lot for the contribution!

This is a little bit more tricky because making the `LocaionComponent` nullable would require changing the annotation in:
https://github.com/mapbox/mapbox-gl-native-android/blob/d36412e131fcb212ea531f1c72fb8e99142bf8e4/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java#L2100-L2103

which in turn will break downstream Kotlin builds. We've decided to treat nullability annotations changes semver-breaking in the past because of this reason. Since we're not planning a major version release soon, I'm afraid this change might have to wait.

Just out of curiosity - there's not a lot of going on in the constructor:
https://github.com/mapbox/mapbox-gl-native-android/blob/d36412e131fcb212ea531f1c72fb8e99142bf8e4/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java#L188-L194

and all of the interactions are dismissed if the component is not activated (initialized):
https://github.com/mapbox/mapbox-gl-native-android/blob/d36412e131fcb212ea531f1c72fb8e99142bf8e4/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java#L1211-L1213

So I'd be interested in what types of performance benefits do the opting out bring? Are there any other operations ongoing that you're noticing when the component is not activated that we might optimize instead?

maxtracking:
@LukasPaczos Oops, I did miss the annotation, thanks for pointing out :)
Our hardware is really lo-spec so eg Raspberry Pi is a space rocket compared to that, not mentioning regular phones. So the fact that `MapView` calls `MapboxMap` on every lifecycle event or when style is loaded is a small but not entirely insignificant bit. I totally agree with you that it's not a lot going on with the component (and I'd gladly remove Telemetry classes instead :) ) . From my manual testing without LocationComponent and telemetry it is faster, not by significant margin but nonetheless. 
So the reasoning is as follows:
1. It is never going to be used for us
2. GPS location is retrieved from external sources
3. Block possible future attempt to enable the component

I would 100% understand if this PR is rejected as "not useful" for general user base.
NB: On a different topic - is there a recommended way to measure rendering performance?

LukasPaczos:
Thanks for context @maxtracking!

I think we could make the component lazily initialized instead, to work around semver. The important thing here is to keep track of the component's lifecycle methods while it's not initialized and it can't do it itself. Those are:
```
LocationComponent#onStart
LocationComponent#onStop
LocationComponent#onDestroy
LocationComponent#onStartLoadingMap
LocationComponent#onFinishLoadingStyle
```

We'd possibly need to inject this state permutation in the constructor and cover it with tests. Would you like to take a stab at this?

> On a different topic - is there a recommended way to measure rendering performance?

You can try the
```
MapboxMap#setOnFpsChangedListener(OnFpsChangedListener)
```
It's far from ideal, but it does give a point of reference.

maxtracking:
Hi @LukasPaczos,
> I think we could make the component lazily initialized instead, to work around semver.

This is a good point indeed, I'll give it a go.

Re `setOnFpsChangedListener()` - well, that was our first point of interest, then we switched to actually subscribing to 'addOnWillStartRenderingFrameListener()` etc to get the diffs. The purpose is to help cartography team optimise the map styles where they are tweaking it (expressions and filters) and interested to know what's the impact on the rendering .

Issue #210, opened 2020-03-02
Release patch of tequila/unicorn: android-v8.6.3 / android-v9.0.1


tobrun:
mapbox-gl-native made a patch release of [v1.2.2](https://github.com/mapbox/mapbox-gl-native/releases/tag/maps-v1.2.2). We need to update `/vendor/mapbox-gl-native` of `release-unicorn/release-tequila` with this version and create a new tag for it (which starts a continuous deployment to bintray). For release tequila (v8.6.3), we need to validate how feasible this change is.

Additionally to the gl-native dependency update, I would like to include:
 - [ ] ~update of telemetry library [#201](https://github.com/mapbox/mapbox-gl-native-android/pull/201)~
 - [x] proguard fix for temetry events https://github.com/mapbox/mapbox-gl-native-android/pull/195 

Anything else?

cc @chloekraw @mapbox/maps-android

tobrun:
We won't be updating the telemetry library this means that the AppTurnStileEvent can still be obfuscated. We can however fix that from our end through our proguard config. @Chaoba can you add a proguard config to keep [these](https://github.com/mapbox/mapbox-events-android/tree/master/libtelemetry/src/main/java/com/mapbox/android/telemetry) files ? and make this change to `release-unicorn` and `release-tequila` branches?

tobrun:
Assigning myself, need to bring some build changes to the release branches. Aiming to release tomorrow.

tobrun:
`release-unicorn` required build changes are moving at https://github.com/mapbox/mapbox-gl-native-android/pull/216. `release-tequila` won't be feasible to bring these changes in. I'm suggesting to release without inclusion of gl-native bump, only the proguard changes. 

cc @chloekraw

chloekraw:
@tobrun `release-tequila` patch without gl-native version bump makes sense 👍

chloekraw:
@tobrun our `release-unicorn` 9.0.x is currently on GL-Native's `release-unicorn`, or 1.0.0. I think we should avoid upgrading minor versions of GL-Native (or any dependencies) in patch releases, so let's upgrade to [1.0.1](https://github.com/mapbox/mapbox-gl-native/releases/tag/maps-v1.0.1). (https://github.com/mapbox/mapbox-gl-native/tree/maps-v1.0.1)

Does that sound good to you?

cc/ @Chaoba

chloekraw:
Done

PR #223, opened 2020-03-06
[doc] Update changelog for v9.0.1

Chaoba:
This PR updates the changelog entries to match the upcoming patch release of v9.0.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Mapbox welcomes participation and contributions from everyone.  If you'd like to do so please see the [`Contributing Guide`](https://github.com/mapbox/mapbox-gl-native/blob/master/CONTRIBUTING.md) first to get started.
 
-## 9.0.1 - March 6, 2020
+## 9.0.1 - March 11, 2020
 [Changes](https://github.com/mapbox/mapbox-gl-native-android/compare/android-v9.0.0...android-v9.0.1) since [Mapbox Maps SDK for Android v9.0.0](https://github.com/mapbox/mapbox-gl-native-android/releases/tag/android-v9.0.0):
 ### Bug fixes
  - Fix Proguard configuration to prevent obfuscation of telemetry event classes. [#195](https://github.com/mapbox/mapbox-gl-native-android/pull/195)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ Mapbox welcomes participation and contributions from everyone.  If you'd like to
 ## 9.0.1 - March 6, 2020
 [Changes](https://github.com/mapbox/mapbox-gl-native-android/compare/android-v9.0.0...android-v9.0.1) since [Mapbox Maps SDK for Android v9.0.0](https://github.com/mapbox/mapbox-gl-native-android/releases/tag/android-v9.0.0):
 ### Bug fixes
- - Fix an issue that could cause a map to fail to load resources. [#16161](https://github.com/mapbox/mapbox-gl-native/pull/16161)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 Mapbox welcomes participation and contributions from everyone.  If you'd like to do so please see the [`Contributing Guide`](https://github.com/mapbox/mapbox-gl-native/blob/master/CONTRIBUTING.md) first to get started.
 
+## 9.0.1 - March 6, 2020
+[Changes](https://github.com/mapbox/mapbox-gl-native-android/compare/android-v9.0.0...android-v9.0.1) since [Mapbox Maps SDK for Android v9.0.0](https://github.com/mapbox/mapbox-gl-native-android/releases/tag/android-v9.0.0):
+### Bug fixes
+ - Fix Proguard configuration to prevent obfuscation of telemetry event classes. [#195](https://github.com/mapbox/mapbox-gl-native-android/pull/195)
+### Other changes
+ - Update mapbox-gl-native to v1.2.2. [#216](https://github.com/mapbox/mapbox-gl-native-android/pull/216)
```
chloekraw: Instead of this entry, we should take the bug fixes in the changelog of GL-Native v1.2.2 and add it to our changelog. 

So delete this entry and section, and under "Bug fixes", add:
```
- Fix a crash that could occur when using styles with `LinePattern`. [#16249](https://github.com/mapbox/mapbox-gl-native/pull/16249)
```

chloekraw:
I think we may want to upgrade to GL-Native v1.0.1 instead v1.2.2 to avoid upgrading minor releases of dependencies in patches. ref: https://github.com/mapbox/mapbox-gl-native-android/issues/210#issuecomment-595565875

Chaoba:
> I think we may want to upgrade to GL-Native v1.0.1 instead v1.2.2 to avoid upgrading minor releases of dependencies in patches. ref: [#210 (comment)](https://github.com/mapbox/mapbox-gl-native-android/issues/210#issuecomment-595565875)

@chloekraw I double checked gl-native, before update to v1.2.2, gl-native is on [release-unicon branch](https://github.com/mapbox/mapbox-gl-native/commits/release-unicorn) which already contains change #16161. So I think we should reconsider which version should be used in this patch release

cc: @pengdev

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 Mapbox welcomes participation and contributions from everyone.  If you'd like to do so please see the [`Contributing Guide`](https://github.com/mapbox/mapbox-gl-native/blob/master/CONTRIBUTING.md) first to get started.
 
+## 9.0.1 - March 6, 2020
+[Changes](https://github.com/mapbox/mapbox-gl-native-android/compare/android-v9.0.0...android-v9.0.1) since [Mapbox Maps SDK for Android v9.0.0](https://github.com/mapbox/mapbox-gl-native-android/releases/tag/android-v9.0.0):
+### Bug fixes
+ - Fix an issue that could cause a map to fail to load resources. [#16161](https://github.com/mapbox/mapbox-gl-native/pull/16161)
```
chloekraw: Add: "on some hardware."

```
 - Fix an issue that could cause a map to fail to load resources on some hardware. [#16161](https://github.com/mapbox/mapbox-gl-native/pull/16161)
```

chloekraw:
Confirmed that we should move forward with upgrading to version 1.0.1 of GL-Native in this patch release.

CHANGELOG.md
```diff
@@ -5,7 +5,6 @@ Mapbox welcomes participation and contributions from everyone.  If you'd like to
 ## 9.0.1 - March 6, 2020
```
chloekraw: nit: update date to today

PR #224, opened 2020-03-06
Revert gl-native to release-unicorn

Chaoba:
This pr revert submodule gl-native to release-unicorn


zugaldia:
Interesting - we got a native crash on the Firebase tests.

PR #278, opened 2020-03-16
eglCreateWindowSurface() can only be called with an instance of Surface

tobrun:
Fixes a crash: "eglCreateWindowSurface() can only be called with an instance of Surface, SurfaceView, SurfaceHolder or SurfaceTexture at the moment"

`<changelog>Fixed a crash that might occur during EGL window surface creation.</changelog>`

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/renderer/textureview/TextureViewRenderThread.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/renderer/textureview/TextureViewRenderThread.java
@@ -1,9 +1,11 @@
 package com.mapbox.mapboxsdk.maps.renderer.textureview;
 
 import android.graphics.SurfaceTexture;
+
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.UiThread;
+
 import android.view.TextureView;
 
 import com.mapbox.mapboxsdk.log.Logger;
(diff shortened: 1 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/renderer/textureview/TextureViewRenderThread.java
```diff
@@ -390,7 +392,7 @@ boolean createSurface() {
 
       // Create an EGL surface we can render into.
       TextureView view = textureViewWeakRef.get();
-      if (view != null) {
+      if (view != null && view.getSurfaceTexture() != null) {
```
chloekraw: LGTM, though noting that Android docs generally recommend using the listener instead to ensure the SurfaceTexture is not just present but also available: 

> A TextureView's SurfaceTexture can be obtained either by invoking `getSurfaceTexture()` or by using a `SurfaceTextureListener`. It is important to know that a SurfaceTexture is available only after the TextureView is attached to a window (and `onAttachedToWindow()` has been invoked.) It is therefore highly recommended you use a listener to be notified when the SurfaceTexture becomes available.

ref: https://developer.android.com/reference/android/view/TextureView

tobrun: We actually do use `SurfaceTextureListener` to obtain and destroy the `TextureSurface` in `TextureViewRenderThread`. The code above is being executed in the EGLHolder, which doesn't have access to wrapping TextureViewRenderThread. In this case we don't need the actual instance of SurfaceTexture, just validating state is enough here.

PR #288, opened 2020-03-23
Add example using within expression

tobrun:
ezgif com-video-to-gif (12)

### DIFF
--- a/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/turf/WithinExpressionActivity.kt
+++ b/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/turf/WithinExpressionActivity.kt
@@ -1,0 +1,268 @@
+package com.mapbox.mapboxsdk.testapp.activity.turf
+
+import android.graphics.Color
+import android.os.Bundle
+import android.os.Handler
+import android.os.PersistableBundle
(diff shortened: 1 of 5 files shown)


alexshalamov:
@tobrun would be cool if bufferLineStringGeometry is visualized, maybe worth adding semitransparent line layer that will draw 'within' geometry?

MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/turf/WithinExpressionActivity.kt
```diff
@@ -0,0 +1,266 @@
...
+      )
+    )
+
+    // Convert to a LineString
+    val routeLineString = LineString.fromLngLats(coordinates)
+
+    // Create buffer around linestring
+    val bufferedRouteGeometry = bufferLineStringGeometry(routeLineString)
+
+    // Setup style with custom layers,
```
chloekraw: Why does this say "custom layers" here? Seems a bit confusing with the `MGLOpenGLStyleLayer` custom layers

tobrun: custom layers on top mapbox streets, I will change the naming ;)

tobrun:
ezgif com-crop

PR #289, opened 2020-03-23
High contrast accessibility mode fix

tobrun:
closes https://github.com/mapbox/mapbox-gl-native-android/issues/31

`<changelog>Fixed a bug that wrongly generated glyphs when high contrast accessibility mode was active by using RGB values instead of alpha channel. /changelog>`

image


chloekraw:
@tobrun @Chaoba would be good to cherrypick this one into `vanillashake`!

PR #298, opened 2020-03-29
Update ccache config

Chaoba:
This pr update ccache config, because there will be compile errors if ccache is not installed or configured.


chloekraw:
cc @tmpsantos @1ec5

PR #304, opened 2020-04-01
Add log for canceled request.

Chaoba:
`<changelog>Add log message for requests canceled  by sdk.</changelog>`
Resolves https://github.com/mapbox/mapbox-gl-native/issues/15810


zugaldia:
@Chaoba @tobrun I'm not sure this solves the original issue. My understanding was that there already was a log message being emitted for cancelled requests. Additionally, we want the error message to offer a better explanation. I suggest something like the following:

> [HTTP] This request was cancelled (%s). This is expected for prefetched tiles that are no longer needed for the map to render.

PR #318, opened 2020-04-10
Not rebuild all files while running test app.

Chaoba:
`<changelog>Update cmake config to let android studio not rebuild all files while running test app.</changelog>`


chloekraw:
Thanks @Chaoba! This is a fantastic improvement. 

Because it's a build change, it doesn't need to be in the changelog.

PR #319, opened 2020-04-10
Expose LocationIndicatorLayer and an alternative rendering mode based on it

LukasPaczos:
This PR exposes a new `LocationComponent`'s rendering mode that can be activated using the options flag:
```
LocationComponentActivationOptions.Builder#useSpecializedLocationLayer
```

The flag defaults to `false`, but when set, it uses a dedicated layer to render the puck instead of using the stack symbol and circle layers. The specialized implementation is fully compatible with the traditional impl, except for:
- Constants like `FOREGROUND_LAYER` or `LOCATION_SOURCE` which are either no-op or return different types in this mode.
- All options that alter the image ID, like `LocationComponentOptions#foregroundName` are ignored. There's a simple alternative in the form of `LocationComponentOptions#foregroundDrawable` and other though.

This branch is currently based on https://github.com/mapbox/mapbox-gl-native/pull/16388 and also requires a fix from https://github.com/mapbox/mapbox-gl-native/pull/16391 to be completely stable. When all those changes are released, I'll retarget the submodule pin.

TODO:
- [x] `IndicatorLocationLayerRenderer` tests.
- [x] Implement location click listener. Because the new layer is not feature-based and we cannot simply query the map, there's no straightforward way to achieve this. We're considering different approaches, like the indicator layer returning its dimensions, so that we can calculate the hitbox. This is TBD and will be resolved or declared incompatible before this lands.
- [x] Changelog.

/cc @chloekraw @galinelle


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentActivationOptions.java
```diff
@@ -208,6 +233,28 @@ public Builder useDefaultLocationEngine(boolean useDefaultLocationEngine) {
...
+     * </ul>
+     *
+     * @param useSpecializedLocationLayer true if you want to initialize and use the
+     *                                    specialized location layer. Defaults to false.
+     * @return the {@link Builder} object being constructed
+     */
+    public Builder useSpecializedLocationLayer(boolean useSpecializedLocationLayer) {
+      this.useSpecializedLocationLayer = useSpecializedLocationLayer;
+      return this;
+    }
```
LukasPaczos, line 256: @chloekraw I'd love your review on the documentation here.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentActivationOptions.java
```diff
@@ -116,6 +120,25 @@ public boolean useDefaultLocationEngine() {
     return useDefaultLocationEngine;
   }
 
+  /**
+   * True if you want to initialize and use the specialized location layer to render the location puck.
+   * <p>
+   * The specialized layer comes with significant performance improvements,
```
chloekraw, line 126:
```suggestion
   * The specialized layer is not based on runtime styling. This brings performance improvements,
   * but also means that it is not backwards compatible. The incompatibilities are:
```

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentActivationOptions.java
```diff
@@ -116,6 +120,25 @@ public boolean useDefaultLocationEngine() {
...
+   * but since it's not based on the runtime styling,
+   * it's not fully compatible with the traditional implementation. The incompatibilities are:
+   * <ul>
+   * <li> Constants like {@link LocationComponentConstants#FOREGROUND_LAYER}
+   * or {@link LocationComponentConstants#LOCATION_SOURCE} are ignored.
+   * <li> All options that alter the image ID, like {@link LocationComponentOptions#foregroundName()}, are ignored.
+   * Use {@link LocationComponentOptions#foregroundDrawable()} instead.
```
chloekraw, line 133:
```suggestion
   * Use {@link LocationComponentOptions#foregroundDrawable()} to alter the image rendered as the puck.
```

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentActivationOptions.java
```diff
@@ -208,6 +233,28 @@ public Builder useDefaultLocationEngine(boolean useDefaultLocationEngine) {
       return this;
     }
 
+    /**
+     * True if you want to initialize and use the specialized location layer to render the location puck.
```
chloekraw, line 237: This is all the same as above, right?

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentActivationOptions.java
chloekraw: Added two suggestions -- otherwise, looks good! Thanks @LukasPaczos.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentActivationOptions.java
LukasPaczos: Yes, it's the same

LukasPaczos:
There are still 2 PRs in flight that I'm waiting for before merging - https://github.com/mapbox/mapbox-gl-native/pull/16400 and https://github.com/mapbox/mapbox-gl-native/pull/16404.

If needed, we could also merge this PR and release an alpha without the location click support that relies on https://github.com/mapbox/mapbox-gl-native/pull/16400 and follow up in the next pre-release.

LukasPaczos:
Changelog entry is below @chloekraw if you want to give it a pass.
```
<changelog>Introduced `LocationComponentActivationOptions#useSpecializedLocationLayer` flag. When enabled, the `LocationComponent` uses a designated map layer to render the location puck which can greatly increase the performance and reduce the rendering overhead. Check the flag documentation for more details on compatibility.</changelog>
```

Otherwise, I bumped all of the deps and I'll merge tomorrow after QA.

LukasPaczos:
Here's a quick comparison between the default rendering (symbol layers):
ezgif com-optimize (19)

and the specialized location layer:
ezgif com-optimize (20)

/cc @mapbox/navigation-android

Issue #197, opened 2020-02-25
Port circle-sort-key


tobrun:
We already have symbol sort keys but now circle/line as well:
 - https://github.com/mapbox/mapbox-gl-native/pull/15875
 - https://github.com/mapbox/mapbox-gl-native/pull/15839

chloekraw:
👋 looking at the [changes needed](https://github.com/mapbox/mapbox-gl-native-android/pull/209/files) for line- and fill-sort-key bindings, I'm interested in pairing with someone on this one after the JNI work is done.

The changes needed look fairly straightforward from what I can see: the main work is on CircleLayer.java and PropertyFactory.java, followed by adding tests. I think this would be a great experience for me to learn more about how to add tests to our SDK when we're implementing new features.

Chaoba:
@chloekraw These changes are made by a script.

 If cycle-sort-key is ready in gl-native, we just need comment out this [line](https://github.com/mapbox/mapbox-gl-native-android/blob/f27b16562d3d15c8a0426134eaf52ad35a3d4c86/scripts/generate-style-code.js#L10)  and run the command `make android-style-code` all changes will be generated automatically.

chloekraw:
@Chaoba oh wow, that's so cool!

PR #322, opened 2020-04-15
Add bindings for circle-sort-key

chloekraw:
`<changelog>Added methods to set and get the sort key of features in a CircleLayer at runtime. Features with a higher sort key will appear above features with a lower sort key.</changelog>`

Fixes https://github.com/mapbox/mapbox-gl-native-android/issues/197


chloekraw:
@LukasPaczos @tobrun I noticed running the script included changes that are also in https://github.com/mapbox/mapbox-gl-native-android/pull/319. Should I manually override the changes that don't relate to circle sort key in a new commit? Let me know what the best way forward is.

chloekraw:
@LukasPaczos thanks! Looks like @Chaoba’s commit removed the non-circle-sort-key changes so hopefully you don’t run into too many merge conflicts when you rebase.

PR #325, opened 2020-04-17
Fix some compass related issues

ystsoi:
This PR fixes some compass related issues, and should be able to solve https://github.com/mapbox/mapbox-gl-native-android/issues/206.

I have also prepared a program to test the compass behavior:
https://github.com/ystsoi/mapbox-gl-native-android/tree/test-compass/TestCompass

There are several changes:
1) Remove the use of the deprecated TYPE_ORIENTATION sensor.
The heading will be wrong if this sensor is used and the device is not in its normal orientation. This can be seen using the test program. But, actually, I do not have any device which will drop to use this sensor, and I think that this sensor should be safe to remove.

2) Prevent dropping the TYPE_ACCELEROMETER and TYPE_MAGNETIC_FIELD sensor events.
The current logic only updates either gravityValues or magneticValues for each 500ms. As the calculation logic requires both gravityValues and magneticValues, and lowPassFilter is involved, so it should be better to process all sensor events, while just update the heading for each 500ms. I only have an old device not having the TYPE_ROTATION_VECTOR sensor, and will use these sensors, so I suppose that most devices will not be affected by this change.

3) Calculate the heading differently, depending on whether the device is placed horizontally or vertically.
The current logic only works well when the device is placed vertically, like for car navigation. But it does not show a reliable heading when the device is placed horizontally. For example, if a group of hikers use the same Mapbox app, they may get different direction info, depending on how they hold their phones. The fix uses the pitch to determine how to calculate the heading. The fix is based on the existing logic, SensorManager API doc, and test results. There is not much documentation on this subject, but you may try the test program to see the effect of the new logic.

4) Update the heading, even if the sensor is unreliable.
The current logic will stop rotating the heading if the sensor is unreliable. To prevent user confusion, developers can either hide the heading, or help to do calibration. The change keeps updating the heading, even if the sensor is unreliable. This makes it possible to use a different, like a gray-out, indicator for the unreliable heading. As sometimes it is difficult to calibrate, the sensor keeps providing unreliable values. Developers can still hide the heading, if they like.

5) Add the mapbox_user_bearing_icon_unreliable drawable for convenience.
The drawable is not used by the SDK. Developers can just switch between mapbox_user_bearing_icon and mapbox_user_bearing_icon_unreliable, and do not need to draw their own icon. mapbox_user_bearing_icon_unreliable is just a gray-out mapbox_user_bearing_icon. See whether it is OK to add it to the SDK.

Please review.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentCompassEngine.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponentCompassEngine.java
@@ -69,16 +69,10 @@ class LocationComponentCompassEngine implements CompassEngine, SensorEventListen
     this.sensorManager = sensorManager;
     compassSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ROTATION_VECTOR);
     if (compassSensor == null) {
-      if (isGyroscopeAvailable()) {
-        Logger.d(TAG, "Rotation vector sensor not supported on device, "
-                + "falling back to orientation.");
-        compassSensor = sensorManager.getDefaultSensor(Sensor.TYPE_ORIENTATION);
-      } else {
-        Logger.d(TAG, "Rotation vector sensor not supported on device, "
-                        + "falling back to accelerometer and magnetic field.");
-        gravitySensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER);
-        magneticFieldSensor = sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD);
-      }
+      Logger.d(TAG, "Rotation vector sensor not supported on device, "
+              + "falling back to accelerometer and magnetic field.");
+      gravitySensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER);
+      magneticFieldSensor = sensorManager.getDefaultSensor(Sensor.TYPE_MAGNETIC_FIELD);
     }
   }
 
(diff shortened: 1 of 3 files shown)


chloekraw:
@ystsoi thank you for this contribution and for clearly documenting your proposed changes. Your suggestions sound reasonable and desirable upon an initial read, but we'll take a closer look and get back to you.

LukasPaczos:
Thanks a lot @ystsoi!

The changes look great, but I have one question - with the current logic when the device is face down, and the angle between the screen and ground is less than 45 degrees, the returned bearing will be flipped by 180 degrees. Could we calculate the value in a way that adjusts for this case?

Let me give you a drawing, and I apologize up-front for the terrible quality:
Screenshot from 2020-04-22 18-12-55

In this example, the user is facing south, leaned back enough that the angle is smaller than 45 degrees, and the compass engine will point north, while I think it should still point south.

LukasPaczos:
@kmadsen since you've been working with sensor data recently, could you also review the changes here?

ystsoi:
Hi @LukasPaczos, for the mentioned case, actually, some compass apps will show north, while some will show south. The updated logic does not handle this case specially, and will need further adjustment if really need this behavior. getOrientation also provides a roll value which may be used to check this case.

ystsoi:
It seems that iOS apps will show north. On Android, Google Maps will show south, while other apps varies. Let me try to adjust it further.

ystsoi:
Handled the remain two cases. No flipping now. Can just comment out the unneeded cases if change mind in the future.

LukasPaczos:
Thanks @ystsoi! Cherry-picking the commits in https://github.com/mapbox/mapbox-gl-native-android/pull/338 to let the tests run.

LukasPaczos:
https://github.com/mapbox/mapbox-gl-native-android/pull/338 landed :tada:

PR #330, opened 2020-04-21
Support rotating the map while tracking

ystsoi:
The PR is the quick fix which I mentioned in https://github.com/mapbox/mapbox-gl-native-android/issues/61.

It adds rotateWhileTracking for rotating the map when a tracking mode is on. And it uses rotateWhileTracking to reset the map to face north during tracking. The rotateWhileTracking logic is similar to those of tiltWhileTracking and zoomWhileTracking, but works on the camera bearing instead.

Please review. And as I built aar and run unit tests via docker, and I have not setup the environment to run the instrumentation tests, so please help to run them too.

There is a test program in
https://github.com/ystsoi/mapbox-gl-native-android/tree/test-rotate-while-tracking/TestRotateWhileTracking, you can use it to test the compass view and resetNorth, if you like. Need to put MapboxGLAndroidSDK-release.aar into app/libs first, and set the access token in MainActivity.java.


ystsoi:
I have setup Android Studio to run the instrumentation tests. They seem OK.

GooPm023:
> The PR is the quick fix which I mentioned in #61.
> 
> It adds rotateWhileTracking for rotating the map when a tracking mode is on. And it uses rotateWhileTracking to reset the map to face north during tracking. The rotateWhileTracking logic is similar to those of tiltWhileTracking and zoomWhileTracking, but works on the camera bearing instead.
> 
> Please review. And as I built aar and run unit tests via docker, and I have not setup the environment to run the instrumentation tests, so please help to run them too.
> 
> There is a test program in
> https://github.com/ystsoi/mapbox-gl-native-android/tree/test-rotate-while-tracking/TestRotateWhileTracking, you can use it to test the compass view and resetNorth, if you like. Need to put MapboxGLAndroidSDK-release.aar into app/libs first, and set the access token in MainActivity.java.

tobrun:
Main branch was renamed from master to main, please retarget and reopen if this PR is still actionable.

ystsoi:
It seems that android-v9.6.0-alpha.1 still has not solved #61.

I created this PR, because I thought that this might help to solve the issue faster. I didn't expect that I need to maintain this PR for nearly half a year. So I think that it is better for me to consider this PR as rejected, so as to release myself.

Maybe, you work on the issue yourselves later, when you have time. The issue is not critical anyway. But according to the Mapbox Terms of Service, it seems that we cannot use a custom build of the SDK for apps using the Mapbox services. So we still need to wait for your fix.

PR #334, opened 2020-04-22
Cherry pick 9.2.0-alpha1

Chaoba:
This pr cherry pick gl-native and changelog for 9.2.0-alpha1.


tobrun:
alpha releases don't require a release branch yet. We directly cut them from master. We only create a release branch as part of the beta release. This is to minimize overhead of cherry-picking like above and upcoming changes next week before the beta.

PR #340, opened 2020-04-24
Fix fast fling issue on devices with low-density screen

ystsoi:
I found that fling gesture will cause a larger movement on devices with lower density screen, which makes Mapbox not work well on those devices. By the proposed change in MapGestureDetector, the issue can be fixed. But I am not sure whether this is the appropriate place to fix, it depends on what input transform.moveBy expects. Please review.

I have also tuned some fling related constant. I use a lower VELOCITY_THRESHOLD_IGNORE_FLING to allow some short flings. This makes map pan appear smoother. And it seems not good that even the minimum fling will cause a large distance movement. Also, a longer ANIMATION_DURATION_FLING_BASE is used to prevent fast moving short flings. See whether you agree on the settings.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java
@@ -66,7 +66,7 @@ public class MapboxConstants {
   /**
    * Animation time of a fling gesture
    */
-  public static final long ANIMATION_DURATION_FLING_BASE = ANIMATION_DURATION;
+  public static final long ANIMATION_DURATION_FLING_BASE = 500;
 
   /**
    * Velocity threshold for a fling gesture
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java
@@ -66,12 +66,12 @@ public class MapboxConstants {
   /**
    * Animation time of a fling gesture
    */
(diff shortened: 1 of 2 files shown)


chloekraw:
@LukasPaczos was this issue resolved in https://github.com/mapbox/mapbox-gl-native-android/pull/356? The changelog entry provided in the PR suggests so, but I'm not seeing where the velocity threshold for the fling gesture was adjusted in the commits.

LukasPaczos:
This is unrelated @chloekraw.

chloekraw:
@LukasPaczos kk, I asked because the other PR has a second suggested changelog entry that seemed to relate to this:

> `<changelog>Fix an issue where the `onFling` listener was called even when the velocity of the gesture did not meet the threshold to actually animate the camera</changelog>`

I excluded that entry from the changelog -- should I add it back in?

LukasPaczos:
Yes, this should be added. https://github.com/mapbox/mapbox-gl-native-android/pull/356 just made 2 changes, but both are unrelated to this one.

ystsoi:
Updated to use a longer ANIMATION_DURATION_FLING_BASE, as the map is still moving too fast, and it is not feel very good.

I found that Transform.moveBy will call NativeMap.moveBy. NativeMap is an interface, so probably, NativeMapView.moveBy will be called. NativeMapView.moveBy will call:
```
nativeMoveBy(dx / pixelRatio, dy / pixelRatio, duration);
```
As dx and dy will be adjusted by pixel ratio there, so there is no need to adjust offsetX and offsetY by pixel ratio before calling transform.moveBy.

Can you reproduce the fast fling issue on devices with low-density screen? If no real device, can also use the emulator to check.

Chaoba:
@ystsoi Thanks for your contribution. I have cherry-picked your changes and created a new [PR](https://github.com/mapbox/mapbox-gl-native-android/pull/503)

ystsoi:
Thanks. Look forward to have it commit. Currently, I need to override several classes to work around the issue.

PR #353, opened 2020-04-29
Re-add license file to repo

zugaldia:
It seems that during the migration from https://github.com/mapbox/mapbox-gl-native (more information: https://github.com/mapbox/mapbox-gl-native/issues/15971) the license file got missing. This PR re-adds the license file as `LICENSE-SDK.md`. To be clear: this is not a change in license, it simply re-adds the missing file to match the previous set up.

The current `LICENSE.md` file does not include the SDK license, but the license of its dependencies.

/cc: @mapbox/maps-android

### DIFF
--- a/LICENSE-SDK.md
+++ b/LICENSE-SDK.md
@@ -1,0 +1,9 @@
+mapbox-gl-native-android Copyright 2014-2020 Mapbox
+
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+
+2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+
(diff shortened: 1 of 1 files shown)


chloekraw:
@zugaldia on iOS, this license appears at the top of LICENSE.md. I think that might be clearer for developers, particularly since when you click on the "View license" tab on GitHub, it takes you to LICENSE.md. What do you think about doing the same for this repo? https://github.com/mapbox/mapbox-gl-native-ios/edit/master/LICENSE.md 

The one consideration is that that means we may have to add this to @tobrun's license generation script

LukasPaczos:
To address @chloekraw comment, the same text could be added to the https://github.com/mapbox/mapbox-gl-native-android/blob/master/scripts/generate-license.py, similarly to the
https://github.com/mapbox/mapbox-gl-native-android/blob/859f64f97ee7165bcacce82f31c2da9be1e3a0ba/scripts/generate-license.py#L13-L25
which will make it appear in the `LICENSE.md`.

chloekraw:
Thank you @LukasPaczos, I'll make this change and run the script for the beta.

PR #368, opened 2020-05-06
Rollback to System.loadLibrary to load native library.

### LINKED ISSUE #367: Investigate using system library loader
The Maps SDK replaced `System.loadLibrary` with Relinker based on a recommendation from NDK 19's changelog entry:
> ndk-depends has been removed. We believe that ReLinker is a better solution to native library loading issues on old Android versions. 

We did so, and later replaced Relinker with Soloader due to stability issues. However, quite a few issues reported have been linked to Soloader as well. We should investigate returning to `System.loadLibrary`, particularly since the need to support "old Android versions" is less today.

cc @mapbox/maps-android @zugaldia

Chaoba:
`<changelog>Fix loading library failure by rollbacking to System.loadLibrary if SoLoader failed to load the library.
</changelog>`
Resolves #367

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/loader/LibraryLoaderProviderImpl.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/loader/LibraryLoaderProviderImpl.java
@@ -2,16 +2,10 @@ package com.mapbox.mapboxsdk.module.loader;
 
 import com.mapbox.mapboxsdk.LibraryLoader;
 import com.mapbox.mapboxsdk.LibraryLoaderProvider;
-import com.mapbox.mapboxsdk.Mapbox;
-import com.mapbox.mapboxsdk.exceptions.MapboxConfigurationException;
-
-import static com.facebook.soloader.SoLoader.init;
-import static com.facebook.soloader.SoLoader.loadLibrary;
 
 /**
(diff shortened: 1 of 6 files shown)


chloekraw:
Oh, this is a nice idea! I was originally thinking that we'd need to do some profiling around Soloader vs the system library, but I think the way this is implemented, it would be safe to cherrypick into `release-water` and hopefully reduce crash rates. @tobrun, would you agree?

cc @mapbox/maps-android

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/module/loader/LibraryLoaderProviderImpl.java
```diff
@@ -41,9 +40,8 @@ public void load(String name) {
         init(Mapbox.getApplicationContext(), false);
         loadLibrary(name);
       } catch (MapboxConfigurationException exception) {
-        Logger.e(TAG, "Couldn't load so file with relinker, application context missing, "
-          + "call Mapbox.getInstance(Context context, String accessToken) first");
-        throw new UnsatisfiedLinkError(exception.getMessage());
+        //Rollback to System.loadLibrary if SoLoader failed to load library.
+        System.loadLibrary(name);
```
tobrun, line 44: This code block won't be hit when SoLoader fails to load (we would hit a UnsatisfiedLinkError not a MapboxConfgurationException)

PR #373, opened 2020-05-08
Fix text cut off when halo text width is applied

Chaoba:
`<changelog>Fix the issue that top of the text is cut off when halo text width is applied.</changelog>`

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
@@ -31,7 +31,7 @@ public class LocalGlyphRasterizer {
 
     paint = new Paint();
     paint.setAntiAlias(true);
-    paint.setTextSize(24);
+    paint.setTextSize(22);
 
     canvas = new Canvas();
     canvas.setBitmap(bitmap);
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
@@ -31,7 +31,7 @@ public class LocalGlyphRasterizer {
 
     paint = new Paint();
     paint.setAntiAlias(true);
-    paint.setTextSize(22);
+    paint.setTextSize(24);
 
     canvas = new Canvas();
     canvas.setBitmap(bitmap);
@@ -52,7 +52,7 @@ public class LocalGlyphRasterizer {
   protected Bitmap drawGlyphBitmap(String fontFamily, boolean bold, char glyphID) {
     paint.setTypeface(Typeface.create(fontFamily, bold ? Typeface.BOLD : Typeface.NORMAL));
     canvas.drawColor(Color.WHITE);
-    canvas.drawText(String.valueOf(glyphID), 0, 20, paint);
+    canvas.drawText(String.valueOf(glyphID), 5, 25, paint);
     return bitmap;
   }
 }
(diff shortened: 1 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
```diff
@@ -31,7 +31,7 @@ of the bitmap (y: 20) with some buffer around the edge
 
     paint = new Paint();
     paint.setAntiAlias(true);
-    paint.setTextSize(24);
+    paint.setTextSize(22);
```
tsuz, line 34: Will this cause an issue if a developer sets a larger font size via the app or via studio?

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
```diff
@@ -31,7 +31,7 @@ of the bitmap (y: 20) with some buffer around the edge
 
     paint = new Paint();
     paint.setAntiAlias(true);
-    paint.setTextSize(24);
```
alexshalamov, line 34: @Chaoba All locally and server generated fonts are rendered at 24 points. Maybe you need to increase bitmap size. Issue needs bit more investigation, 24pt glyph should fit into 35x35 bitmap.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
Chaoba: What's the text size do you set in your style? Can you set it to a very large one to help me verify it?

alexshalamov: @tsuz @Chaoba `24pt` is a base font size across whole mapbox ecosystem, server and client side. This value should not be changed.

We use SDF rasterezed glyphs, they can be up/downscaled without much artifacts. I don't see how setting very large text size in studio would help here.

I think the issue is that for some reason, Android renders CJK glyph too close to a bitmap border (we expect to have have 3px padding around rendered glyph)
35x35px buffer comes from 24px + 3px border * 2 + 1px glyph padding * 2 + safe margin.

@Chaoba could you increase buffer size and see if truncation is fixed? You can also save bitmap locally and see if there is a truncation in SDF glyph.

Chaoba:
Save some Chinese bitmap and we can find all characters are on the left top
image

And for `Noto Sans CJK JP Bold`, the characters are on left top too and some are cut off
image

We can move characters a litter right and bottom to fix this issue. @alexshalamov What do you think?

Chaoba:
Characters will be in the center after modification. 
image

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
Chaoba: Increate bitmap size will twist characters.

chloekraw:
If this is possible to resolve by the same time as https://github.com/mapbox/mapbox-gl-native-android/issues/377 and we feel good about the resolution, let's cherrypick this into v9.2.1 as well.

Chaoba:
The current solution will cause irregular while text contains numbers.
image

alexshalamov:
> The current solution will cause irregular while text contains numbers.

@Chaoba that is because a glyph was moved within a bitmap and is out-of-sync with glyph metrics, therefore, CJK glyph baseline is 'lower' than latin character's baseline. Do you have a branch with a test that I can try?

Chaoba:
Thanks to @alexshalamov, it looks good now.
image

alexshalamov:
@1ec5 FYI, might be applicable to iOS / darwin code as well.

chloekraw:
> maybe worth adding unit tests to catch regressions

Yes please, @Chaoba could you add a test?

Chaoba:
It's not easy to verify this part by a test case. @tobrun Do you have any suggestions on it?

alexshalamov:
> It's not easy to verify this part by a test case.

@tobrun Are there any examples of 'pixelmatch' tests that verify rendering result?

tobrun:
> @tobrun Are there any examples of 'pixelmatch' tests that verify rendering result?

I had a small POC render test in https://github.com/mapbox/mapbox-gl-native/pull/14687 but that became obsolete with the platform specific c++ test-runner. This PR however highlights that we need to support platform specific pixelmatch testing. I will put some notes down into a separate issue for an Android instrumented unit test integration 

cc @julianrex

PR #378, opened 2020-05-09
Update location while camera move in tracking mode

### LINKED ISSUE #44: [Android] Double tapping while tracking zooms to location, not puck
Double tapping the map while tracking the user causes the camera to zoom to the tapped location, while the camera mode is still "tracking" the user's puck. This behavior will look weird for the user, since it won't look like the puck is still being tracked (although, at location updates, the camera will start actually tracking again).

In previous versions (I'm afraid I don't know the exact versions) the camera would zoom in on the user's puck if the camera was tracking, regardless of where on the map was double tapped.

The zoom behavior seems to work as expected when the user is getting location updates and on the move, and so is likely mostly a problem while standing still.

### Steps to reproduce

 1. Have an idle location
 2. Set camera mode to `CameraMode.TRACKING_GPS_NORTH` or `CameraMode.TRACKING_GPS`
 3. Double tap on the map, somewhere that is not on the user's location

### Expected behavior
The map would zoom in on the user's puck, and the camera would still be tracking the user.

### Actual behavior
The map zooms in on the chosen location, and it looks like the puck is no longer being tracked, even though it is.

I have noticed sometimes that the zoom behavior works correctly when starting up the app, but after moving the map (setting camera mode to non-tracking) and setting camera mode to Tracking again, the issue reappears.

### Configuration

**Mapbox SDK versions:** 8.4.0

Chaoba:
`<changelog>Fix double-tapping while tracking zooms to puck.</changelog>`
#Resolves #44

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
@@ -76,6 +76,7 @@ final class LocationCameraController {
                            AndroidGesturesManager initialGesturesManager,
                            AndroidGesturesManager internalGesturesManager) {
     this.mapboxMap = mapboxMap;
+    mapboxMap.addOnCameraMoveListener(onCameraMoveListener);
     this.transform = transform;
     this.moveGestureDetector = moveGestureDetector;
     this.internalCameraTrackingChangedListener = internalCameraTrackingChangedListener;
@@ -352,7 +353,7 @@ final class LocationCameraController {
 
     @Override
     public void onCameraMove() {
-      if (isLocationTracking() && lastLocation != null) {
+      if (isLocationTracking() && lastLocation != null && options.trackingGesturesManagement()) {
         PointF focalPoint = mapboxMap.getProjection().toScreenLocation(lastLocation);
         mapboxMap.getUiSettings().setFocalPoint(focalPoint);
       }
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
@@ -360,9 +360,9 @@ final class LocationCameraController {
 
     @Override
     public void onCameraMove() {
-      if (isLocationTracking()) {
-        adjustFocalPoint = true;
-        setLatLng(lastLocation);
+      if (isLocationTracking() && lastLocation != null) {
+        PointF focalPoint = mapboxMap.getProjection().toScreenLocation(lastLocation);
+        mapboxMap.getUiSettings().setFocalPoint(focalPoint);
       }
     }
   };
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
@@ -35,7 +35,6 @@ final class LocationCameraController {
   private final Transform transform;
   private final OnCameraTrackingChangedListener internalCameraTrackingChangedListener;
   private LocationComponentOptions options;
-  private boolean adjustFocalPoint;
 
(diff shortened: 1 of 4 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
```diff
@@ -355,6 +356,17 @@ private void notifyCameraTrackingChangeListener(boolean wasTracking) {
     }
   }
 
+  private MapboxMap.OnCameraMoveListener onCameraMoveListener = new MapboxMap.OnCameraMoveListener() {
+
+    @Override
+    public void onCameraMove() {
+      if (isLocationTracking()) {
+        adjustFocalPoint = true;
```
LukasPaczos, line 364: Could the logic that resets the focal point be moved to the camera move listener entirely? This would allow us to clean up the class and not maintain the `adjustFocalPoint` flag and the `lastLocation` object.

tobrun:
cc @pengdev

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
```diff
@@ -355,6 +356,17 @@ private void notifyCameraTrackingChangeListener(boolean wasTracking) {
     }
   }
 
+  private MapboxMap.OnCameraMoveListener onCameraMoveListener = new MapboxMap.OnCameraMoveListener() {
+
+    @Override
+    public void onCameraMove() {
```
LukasPaczos, line 362: Looking at https://github.com/mapbox/mapbox-gl-native-android/blob/de5003813e4bf9d88d6d9eb0980c2ec60f0006bb/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java#L112-L116

the `onCameraMove` will never get called for `moveTo` (which camera controller calls) and it needs to be fixed. If that's handled, we won't need the logic that `adjustFocalPoint` guards in this class, it would be automatically executed with `onCameraMove` listener.

LukasPaczos: Moving `invalidateCameraPosition();` before `cameraChangeDispatcher.onCameraIdle(); ` should resolve the original issue as far as I can tell.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
```diff
@@ -355,6 +348,17 @@ private void notifyCameraTrackingChangeListener(boolean wasTracking) {
...
+      if (isLocationTracking() && lastLocation != null) {
```
LukasPaczos, line 355: missing `options.trackingGesturesManagement()` check

MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationCameraControllerTest.java
```diff
@@ -283,21 +283,35 @@ public void onNewLatLngValue_cameraModeNoneIgnored() {
   @Test
   public void onNewLatLngValue_focalPointIsAdjusted() {
     MapboxMap mapboxMap = mock(MapboxMap.class);
-    UiSettings uiSettings = mock(UiSettings.class);
-    when(mapboxMap.getUiSettings()).thenReturn(uiSettings);
+    Transform transform = mock(Transform.class);
+
+    final MapboxMap.OnCameraMoveListener[] listener = {null};
```
LukasPaczos, line 288: NIT: this doesn't have to be an array, we are always assigning the first element anyway

Chaoba: `listener` is accessed from inner class, this is what lint suggests.

LukasPaczos: :+1:

PR #380, opened 2020-05-12
Fix getFilter with distance expression crash.

Chaoba:
`<changelog>Fixed a crash while Layer.getFilter() has expression distance content. </changelog>`

Update getFilter test for expression within and distance.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
@@ -9,6 +9,10 @@ import com.google.gson.JsonNull;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
 import com.mapbox.geojson.GeoJson;
+import com.mapbox.geojson.LineString;
+import com.mapbox.geojson.MultiLineString;
+import com.mapbox.geojson.MultiPoint;
+import com.mapbox.geojson.Point;
 import com.mapbox.geojson.Polygon;
 import com.mapbox.mapboxsdk.style.layers.PropertyFactory;
 import com.mapbox.mapboxsdk.style.layers.PropertyValue;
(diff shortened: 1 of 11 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
```diff
@@ -4767,6 +4771,18 @@ public static Expression convert(@NonNull JsonArray jsonArray) {
...
+        String json = jsonArray.get(1).toString();
+        if ("Point".equals(type)) {
+          return distance(Point.fromJson(json));
+        } else if ("LineString".equals(type)) {
+          return distance(LineString.fromJson(json));
+        } else if ("MultiPoint".equals(type)) {
+          return distance(MultiPoint.fromJson(json));
+        } else {
+          return distance(MultiLineString.fromJson(json));
+        }
```
zmiao, line 4785: Now `distance` also supports `Polygon` and `MultiPolygon`, so it is better to include these two types also.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/expressions/Expression.java
```diff
@@ -4767,6 +4773,22 @@ public static Expression convert(@NonNull JsonArray jsonArray) {
...
+        } else if ("LineString".equals(type)) {
+          return distance(LineString.fromJson(json));
+        } else if ("MultiPoint".equals(type)) {
+          return distance(MultiPoint.fromJson(json));
+        } else if ("Polygon".equals(type)) {
+          return distance(Polygon.fromJson(json));
+        } else if ("MultiPolygon".equals(type)) {
+          return distance(MultiPolygon.fromJson(json));
+        } else {
+          return distance(MultiLineString.fromJson(json));
```
tobrun, line 4790: We are doing a lot of manual type checking and this logic should be handle, if possible, by the geojson library itself. 
Have you tried using `GeometryGeoJson.fromJson()`?

PR #408, opened 2020-06-01
[ci] Update CI image to use android-ndk-21b.

pengdev:
This PR updates CI image to use android-ndk-21b.


build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.0'
+        classpath 'com.android.tools.build:gradle:3.6.3'
```
tobrun, line 9: can we do 4.0 instead? AS 4.0 stable was released last week

pengdev: 4.0.0 introduced some breaking change, which need to be resolved. I can try to resolve the issues.

pengdev: It seems there is an incompatibility issue with the `android-sdk-versions-plugin` and gradle plugin 4.0.0, the error is printed as follows:
```
Cannot cast object 'task ':MapboxGLAndroidSDK:generateDebugBuildConfig' property 'versionCode'' with class 'org.gradle.api.internal.provider.DefaultProperty' to class 'java.lang.Integer'
```
cc @harvsu

PR #412, opened 2020-06-02
Fix remove source memory leak.

### LINKED ISSUE #405: Memory leak in native code
Mapbox: 9.2.0
Reproducer: https://gist.github.com/Dkhusainov/4cc6605692961c6c075196668a1d54c2

Open AS profiler while running the reproducer and see this picture:
leak

Chaoba:
`<changelog>Fix the memory leak while adding and removing sources.</changelog>`

#Resolve #405


tobrun:
CI fails with 

> Abort message: '/src/MapboxGLAndroidSDK/src/cpp/style/sources/custom_geometry_source.cpp:113: void mbgl::android::CustomGeometrySource::releaseThreads(): assertion "javaPeer" failed'

This is because CustomGeometrySource is a bit different Source as the others and we might need to find a separate solution for this one.

PR #418, opened 2020-06-03
Keep strong reference to background scheduler in geojson_source

tobrun:
`<changelog>Introduced strong reference to background scheduler in geojson source.</changelog>`

Capturing from @alexshalamov :

> Actor holds reference to scheduler (Scheduler&). If non-owned source would be converting feature, background scheduler could be destructed if map instance is destructed. I would rather keep strong reference to background scheduler.


tobrun:
current solution was already safe:

```
 template <class... Args>
    Actor(std::shared_ptr<Scheduler> scheduler, Args&&... args)
        : retainer(std::move(scheduler)), target(*retainer, parent, std::forward<Args>(args)...) {}
```

PR #426, opened 2020-06-11
Rotate camera in the shortest path.

Chaoba:
`<changelog>Fix puck camera rotate.</changelog>`


Chaoba:
@LukasPaczos By the additional reproduce info, I can reproduce it locally. The root cause for this issue is the animator changes from 0 to 360 or 0 to 355(or others).

PR #429, opened 2020-06-15
Not change file source status while changing cache path.

Chaoba:
` fileSource.deactivate()` will pause `MainResourceLoader` and so that it will not work for loading local resources.

This PR removes the process of changing fileSource status to keep `MainResourceLoader` running.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/storage/FileSource.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/storage/FileSource.java
@@ -329,7 +329,6 @@ public class FileSource {
     fileSource.setResourceCachePath(path, new ResourcesCachePathChangeCallback() {
       @Override
       public void onSuccess(@NonNull String path) {
-        fileSource.deactivate();
         resourcesCachePathLoaderLock.lock();
         resourcesCachePath = path;
         resourcesCachePathLoaderLock.unlock();
@@ -342,7 +341,6 @@ public class FileSource {
         callback.onError(message);
       }
     });
-    fileSource.activate();
   }
 
   private static boolean isPathWritable(String path) {
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/storage/FileSource.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/storage/FileSource.java
@@ -293,12 +293,7 @@ public class FileSource {
     final Context applicationContext = Mapbox.getApplicationContext();
     final FileSource fileSource = FileSource.getInstance(applicationContext);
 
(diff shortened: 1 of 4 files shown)


tobrun:
@LukasPaczos as original implementer of this API you have a bit more background on the need to activate/deactivate the FileSource. Could you add some context?

LukasPaczos:
We need to manually activate the file source here, at least we had to in the past. The native `setResourceCachePath` switches the calling thread to the `FileSourceThread` and resets the path and the database on that thread. If it would have been paused, the callback would never return.

That's why we need to activate the file source (and unpause the thread) to complete the operation and later do the cleanup and deactivate to keep the activation counter honest.

There's the [ChangeResourcesCachePathActivity](https://github.com/mapbox/mapbox-gl-native-android/blob/master/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/offline/ChangeResourcesCachePathActivity.kt) example that should reproduce the issue on this branch and never be able to change the path.

Chaoba:
> We need to manually activate the file source here, at least we had to in the past. The native `setResourceCachePath` switches the calling thread to the `FileSourceThread` and resets the path and the database on that thread. If it would have been paused, the callback would never return.

Do a test by deactivating filesource before changing cache path and it could work.

What confuses me is that when a FileSouce instance is created, the `resourceLoader` inside is in activated state while `FileSouce.isActivated()` return false.

If we activate the FileSouce before changing cache path and then deactivate it, `resourceLoader` will be in paused state, that's the reason why it will not load resource after changing cache path.

LukasPaczos:
> the resourceLoader inside is in activated state while FileSouce.isActivated() return false

Maybe the activation counter in the `file_source.cpp` is incorrectly initialized after some recent changes?

Chaoba:
@LukasPaczos Yeah, it is. Init the counter in the constructor method and the default state for `FileSource` is active now.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/storage/FileSource.java
```diff
@@ -293,12 +293,7 @@ public static void setResourcesCachePath(@NonNull final String path,
     final Context applicationContext = Mapbox.getApplicationContext();
     final FileSource fileSource = FileSource.getInstance(applicationContext);
 
-    if (fileSource.isActivated()) {
```
LukasPaczos, line 296: What happens now if we would try to change the cache path while we're processing tiles? What happens if the map is rendering and we call this method, are we going to crash or will we be able to safely reset the database pointer?

Chaoba: Not seeing anything wrong during my test. @alexshalamov for visibility.

PR #437, opened 2020-06-22
Smooth puck animation after changing style.

### LINKED ISSUE #431: Puck jumps on style change
ezgif com-optimize (39)

As seen on the gif, when the map style is changed, the puck jumps forward instead of being animated smoothly. It's understandable that we need to stop the animation for the duration of the style change, but maybe there's a way to resume it from the previous position instead of resetting to the current one.

/cc @tobrun @Chaoba

Chaoba:
Resolves #431 

If puck animation is interrupted by changing style, the puck animation will be stopped at its current position and resumed from its target position. So it looks like puck jumps to the new position. This pr smooth puck animation by recording its breakpoint and resume from this point. 

ezgif com-resize

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
@@ -502,7 +502,6 @@ final class LocationAnimatorCoordinator {
       animator.cancel();
       animator.removeAllUpdateListeners();
       animator.removeAllListeners();
-      animatorArray.put(animatorType, null);
     }
   }
 
--- a/MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinatorTest.kt
+++ b/MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinatorTest.kt
@@ -14,8 +14,7 @@ import com.mapbox.mapboxsdk.location.modes.RenderMode
 import com.mapbox.mapboxsdk.maps.MapboxMap
 import com.mapbox.mapboxsdk.maps.Projection
 import io.mockk.*
-import junit.framework.Assert.assertEquals
-import junit.framework.Assert.assertTrue
+import junit.framework.Assert.*
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -462,7 +461,7 @@ class LocationAnimatorCoordinatorTest {
     locationAnimatorCoordinator.feedNewLocation(Location(""), cameraPosition, true)
     locationAnimatorCoordinator.cancelAllAnimations()
 
-    assertTrue(locationAnimatorCoordinator.animatorArray[ANIMATOR_CAMERA_LATLNG] == null)
+    assertFalse(locationAnimatorCoordinator.animatorArray[ANIMATOR_CAMERA_LATLNG].isRunning)
   }
 
   @Test
@@ -475,7 +474,7 @@ class LocationAnimatorCoordinatorTest {
     )
     locationAnimatorCoordinator.cancelZoomAnimation()
 
-    assertTrue(locationAnimatorCoordinator.animatorArray[ANIMATOR_ZOOM] == null)
+    assertFalse(locationAnimatorCoordinator.animatorArray[ANIMATOR_ZOOM].isRunning)
   }
 
   @Test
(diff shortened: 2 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
```diff
@@ -222,13 +222,17 @@ void feedNewTilt(double targetTilt, @NonNull CameraPosition currentCameraPositio
...
     LatLng previousLatLng;
     MapboxAnimator latLngAnimator = animatorArray.get(ANIMATOR_LAYER_LATLNG);
     if (latLngAnimator != null) {
       previousLatLng = (LatLng) latLngAnimator.getAnimatedValue();
     } else {
-      previousLatLng = new LatLng(previousLocation);
+      if (previousLocation != null) {
+        previousLatLng = new LatLng(previousLocation);
+      } else {
+        previousLatLng = new LatLng(0f, 0f);
```
tobrun, line 234: potentially this could result in a weird flickering, the location jumping from one part of the world to current location. Can we use the current position instead? This will result in a no-op animation but I feel that would be a better UX.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1286,6 +1286,14 @@ private void onLocationLayerStop() {
     }
 
     stopPulsingLocationCircle();
+
+    //Record the current layer location so that location layer animation continues from this point after change style.
+    LatLng currentLayerLocation = locationAnimatorCoordinator.getPreviousLayerLatLng();
+    if (currentLayerLocation != null && lastLocation != null) {
+      lastLocation.setLatitude(currentLayerLocation.getLatitude());
+      lastLocation.setLongitude(currentLayerLocation.getLongitude());
```
LukasPaczos, line 1294: The `lastLocation` field is also used as a return value for the `getLastLocation` that can be called from the outside and should return the true value, not a place where the animation stopped.

MapboxGLAndroidSDK/src/test/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinatorTest.kt
```diff
@@ -462,7 +461,7 @@ class LocationAnimatorCoordinatorTest {
     locationAnimatorCoordinator.feedNewLocation(Location(""), cameraPosition, true)
     locationAnimatorCoordinator.cancelAllAnimations()
 
-    assertTrue(locationAnimatorCoordinator.animatorArray[ANIMATOR_CAMERA_LATLNG] == null)
+    assertFalse(locationAnimatorCoordinator.animatorArray[ANIMATOR_CAMERA_LATLNG].isRunning)
```
LukasPaczos, line 464:
```suggestion
    assertFalse(locationAnimatorCoordinator.animatorArray[ANIMATOR_CAMERA_LATLNG].isStarted)
```

Does this test actually verify the behavior? The animator objects in the array are mocked, so they might always default to false, regardless of what happens before. Could you check whether that's the case?

Chaoba: Oh, right, let's verify these animators are canceled.

Chaoba: Any suggestions on how to verify it? 
Using `verify{locationAnimatorCoordinator.animatorArray[ANIMATOR_LAYER_LATLNG].cancel()}` will cause 
`io.mockk.MockKException: Missing calls inside every { ... } block.` error.

LukasPaczos: This got me thinking, and actually, the animators in the class are not mocks, they are real objects. That's why you cannot verify them. However, they are not started either, because starting happens in a mocked `animatorSetProvider` so just checking `isStarted` flag is not enough because it's always `false`. The test class would need a bigger refactor to remove the dependency on the real animator objects, so I pushed https://github.com/mapbox/mapbox-gl-native-android/pull/437/commits/94d24a1bfe2df309d17d1e3c6652e43e1bf048db as a stopgap. Feel free to remove it and refactor or leave as is.

LukasPaczos: Looking even more at this, I also found a way to this with the current, real object setup. Check out https://github.com/mapbox/mapbox-gl-native-android/pull/437/commits/b58939ffe68eafc0ce2895a775a533d244796d73 and again, feel free to pick one the option that you think works best.

Chaoba: It looks good 👍 . Thank you.

PR #440, opened 2020-06-24
Not changing location camera mode while disabled.

### LINKED ISSUE #291: LocationComponent displayes broken updates even when not enabled
```
mapboxMap.locationComponent.activateLocationComponent(
  LocationComponentActivationOptions.builder(this, style)
    .useDefaultLocationEngine(false)
    .build()
)
mapboxMap.locationComponent.cameraMode = CameraMode.TRACKING_GPS_NORTH

{ // location samples loop
mapboxMap.locationComponent.forceLocationUpdate(enhancedLocation)
}
```

When running the above scenario, the puck will start displaying and updating only the accuracy layer of the puck (changing its position as well) and also track the location with the camera, even though the component was never enabled, like:
```
mapboxMap.locationComponent.isLocationComponentEnabled = true
```

Until the above enable method is called, the component should ignore all updates.

Chaoba:
Resolves #291

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
@@ -610,6 +610,9 @@ public final class LocationComponent {
                             @Nullable Double zoom, @Nullable Double bearing, @Nullable Double tilt,
                             @Nullable OnLocationCameraTransitionListener transitionListener) {
     checkActivationState();
+    if (!isEnabled) {
+      return;
+    }
     locationCameraController.setCameraMode(
       cameraMode, lastLocation, transitionDuration, zoom, bearing, tilt,
       new CameraTransitionListener(transitionListener));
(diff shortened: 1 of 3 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -610,6 +610,9 @@ public void setCameraMode(@CameraMode.Mode int cameraMode,
                             @Nullable Double zoom, @Nullable Double bearing, @Nullable Double tilt,
                             @Nullable OnLocationCameraTransitionListener transitionListener) {
     checkActivationState();
+    if (!isEnabled) {
+      return;
+    }
```
LukasPaczos, line 615: I think this can break some of the existing implementations. If someone was setting the camera mode before enabling the component, that logic will not work anymore. Is there another way that we can resolve the issue?

Chaoba: Then we might need to save the enable state in `LocationCameraController` and only save its camera mode without update camera when disabled.

LukasPaczos:
What happens if we run:
```
activateLocationComponent
setRenderMode(GPS)
setLocationComponentEnabled
```

is the GPS render mode going to be applied?

Chaoba:
@LukasPaczos Yes, the render mode will be saved in `LocationCameraController#setCameraMode`. What we do in this pr is only skip `transitionToCurrentLocation` while in disabled state.

PR #449, opened 2020-07-02
[build] use relative path for retrieving /gradle

tobrun:
When building out a multi folder Android project. The rootDir will resolve to the rootdir of the main project and not the rootdir of the sub project. To consume this repo as submodule in a repo, we need to be explicit about where to find gradle files.


MapboxGLAndroidSDKTestApp/build.gradle
```diff
@@ -1,5 +1,5 @@
 apply plugin: 'com.android.application'
-apply from: "${rootDir}/gradle/native-build.gradle"
+apply from: "../gradle/native-build.gradle"
```
pengdev, line 2: could be `{projectDir}/../`? Even though shouldn't be any differece?

tobrun: that works as well

MapboxGLAndroidSDK/build.gradle
```diff
@@ -96,10 +96,10 @@ configurations {
...
 }
-apply from: "${rootDir}/gradle/gradle-javadoc.gradle"
-apply from: "${rootDir}/gradle/gradle-checkstyle.gradle"
-apply from: "${rootDir}/gradle/gradle-dependencies-graph.gradle"
-apply from: "${rootDir}/gradle/gradle-update-vendor-modules.gradle"
-apply from: "${rootDir}/gradle/android-nitpick.gradle"
-apply from: "${rootDir}/gradle/gradle-bintray.gradle"
-apply from: "${rootDir}/gradle/jacoco-report.gradle"
\ No newline at end of file
+apply from: "../gradle/gradle-javadoc.gradle"
```
pengdev, line 99: And same for the `MapboxGLAndroidSDK/build.gradle`

PR #465, opened 2020-07-09
[core] update gl-native binary to 1.8.1

tobrun:
This PR updates the gl-native binary to v1.8.1 and updates the changelog for v9.3.0-beta.1 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [`Contributing Guide`](https://github.com/mapbox/mapbox-gl-native/blob/master/CONTRIBUTING.md) to get started.
 
+## 9.3.0-beta.1 - July 9, 2020
+ - Updated Mapbox GL Native library to v1.8.1. ([#464](https://github.com/mapbox/mapbox-gl-native-android/pull/464))
+
 ## 9.3.0-alpha.1 - July 2, 2020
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [`Contributing Guide`](https://github.com/mapbox/mapbox-gl-native/blob/master/CONTRIBUTING.md) to get started.
 
+## 9.3.0-beta.1 - July 9, 2020
```
chloekraw, line 5: @mapbox/maps-android reminder that [changelog entries](https://github.com/mapbox/mapbox-maps-internal/blob/master/docs/guidelines-changelog.md) on CHANGELOG.md no longer have separate entries for pre-releases. 

When updating CHANGELOG.md for the beta and stable release, add to the existing changelog from the last pre-release and update the heading, so that at the end of a release cycle, **there is one entry per minor release**. Patch releases are separate entries.

What we have is fine in this case and can be addressed in the next release (beta.2 or stable). 

There should be some way for developers who do want to know what's in each pre-release (e.g., the Nav SDK team) to find that out. We used the [GitHub releases page](https://github.com/mapbox/mapbox-gl-native-android/releases) for this purpose -- each release tag description only contains the changes for that release. Let's discuss whether we'll keep this practice or replace it with something else.

PR #470, opened 2020-07-13
Set gps bearing immediately while changing to gps mode.

### LINKED ISSUE #462: GPS puck keeps previous value when enabled
ezgif com-optimize (46)

When we're not in render mode GPS, the layer renderer ignores GPS bearing updates so when the mode is engaged, it needs to catch up. I believe we can still avoid updating the layer itself with an unnecessary value, but we should cache that value and set it right away when the mode is enabled instead of waiting for a new value to come.

/cc @tobrun @Chaoba

Chaoba:
Cache gps bearing while in normal and compass mode and then set it to render when changing to GPS mode.
Resolves #462

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationLayerController.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationLayerController.java
@@ -106,6 +106,10 @@ final class LocationLayerController {
     }
   }
 
+  void setGpsBearing(float gpsBearing) {
+    locationLayerRenderer.setGpsBearing(gpsBearing);
+  }
+
   void setRenderMode(@RenderMode.Mode int renderMode) {
     if (this.renderMode == renderMode) {
       return;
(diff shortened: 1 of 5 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationLayerController.java
```diff
@@ -303,10 +313,9 @@ public void onNewAnimationValue(Float newPulseRadiusValue) {
   Set<AnimatorListenerHolder> getAnimationListeners() {
     Set<AnimatorListenerHolder> holders = new HashSet<>();
     holders.add(new AnimatorListenerHolder(MapboxAnimator.ANIMATOR_LAYER_LATLNG, latLngValueListener));
+    holders.add(new AnimatorListenerHolder(MapboxAnimator.ANIMATOR_LAYER_GPS_BEARING, gpsBearingValueListener));
```
LukasPaczos, line 316: This will make animator running and computing values even if we're not consuming them. When switching the mode we're not interested in the animated value but rather the final value, so maybe there's a chance that we can push the bearing when updating the mode instead of caching the animation?

Chaoba:
@LukasPaczos Rethink about it and found a new solution to set lastLocation's bearing while change to GPS mode. Also found if it has `ANIMATOR_LAYER_GPS_BEARING` animator, it will start the animation from this animator value to the new location bearing, so remove this animator to eliminate the catch-up animation.

PR #472, opened 2020-07-15
Add http hook

### LINKED ISSUE #307: Make HttpReqeust error codes public, provide developer hook
When a http request fails to execute. End developers don't have much control over the errorflow and can only rely on a string error message (which is dangerous as changes to the value can break an enduser implementation). To provide these customers with a better user experience we need to make HttpRequest error codes public (make them an IntDef?), improve the definitions and provide an error handler hook into our http stack. 

(note that we are bound to the error code values that gl-native expect but we can wrap and improve without regressing that implementation).

Chaoba:
Resolves #307 
This pr adds Http hook which could let developers get the HTTP response codes


Chaoba:
Then this pr will not be needed, closed.

Issue #467, opened 2020-07-09
Integrate user animation handling


tobrun:
In an upcoming gl-native release, we will have the ability to flag that an platform driven animation is ocrruring (similar to the setGestureInProgresss/isGestureInProgress).

tobrun:
Expose Map::setUserAnimationInProgress()/isUserAnimationInProgress()

LukasPaczos:
For reference, `LocationComponent` integration happened in https://github.com/mapbox/mapbox-gl-native-android/pull/501 and https://github.com/mapbox/mapbox-gl-native-android/pull/504.

PR #491, opened 2020-07-22
Expose animation handling

yuryybk:
This PR resolves mapbox/mapbox-gl-native-android#467

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
@@ -2116,6 +2116,17 @@ public final class MapboxMap {
     return locationComponent;
   }
 
+  /**
+   * Tells the map rendering engine that the animation is currently performed by the
+   * user (e.g. with a `jumpTo()` calls series). It adjusts the engine for the animation use case.
+   * In particular, it brings more stability to symbol placement and rendering.
+   *
+   * @param inProgress Bool representing if user animation is in progress
+   */
+   public void setUserAnimationInProgress(boolean inProgress) {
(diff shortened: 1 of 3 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -2116,6 +2116,17 @@ public LocationComponent getLocationComponent() {
...
+   * Tells the map rendering engine that the animation is currently performed by the
+   * user (e.g. with a `jumpTo()` calls series). It adjusts the engine for the animation use case.
+   * In particular, it brings more stability to symbol placement and rendering.
+   *
+   * @param inProgress Bool representing if user animation is in progress
+   */
+   public void setUserAnimationInProgress(boolean inProgress) {
+     nativeMapView.setUserAnimationInProgress(inProgress);
+   }
+
```
Chaoba, line 2129: Miss `isUserAnimationInProgress` here.

yuryybk: yeah, I've exposed only `setUserAnimationInProgress` in the same way as in Carbon, https://github.com/mapbox/mapbox-maps-android/blob/master/sdk/src/main/java/com/mapbox/maps/MapboxMap.kt

should we expose  `isUserAnimationInProgress` only here? or it is missed in Carbon too?

tobrun: We need both getter as setter, need to double check carbon impl but that should have both afaik

yuryybk: exposed `isUserAnimationInProgress`

PR #495, opened 2020-07-24
[offline] remove world bounds validation when creating an offline region

### LINKED ISSUE #494: Enable offline regions that cross the 180th meridian
There is an error thrown when crossing the 180th meridian with Offline Regions that throws 
`Provided OfflineRegionDefinition doesn't fit the world bounds`

**Expected behavior**
It is possible to enter coordinates with longitudes outside of -180-180 in the offline tile count estimator (see https://docs.mapbox.com/playground/offline-estimator/). Users should be able to download regions that cross the 180th meridian.

We should support the LatLngBounds like [10, -170, -10, -190], however while validating the OfflineRegionDefinition in OfflineManager we do check if the bounds are using wrapped coordinates:

```
private boolean isValidOfflineRegionDefinition(OfflineRegionDefinition definition) {
    return LatLngBounds.world().contains(definition.getBounds());
}
```

And will throw exception if the longitude is outside [-180, 180]. Seems like the Offline Region doesn't support download a single bound across the meridian.

We should adjust this logic to allow for bounds across the meridian and remove the `isValidOfflineRegionDefinition` validation.

cc @pengdev @tobrun

tobrun:
closes #494

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
@@ -12,7 +12,6 @@ import androidx.annotation.UiThread;
 
 import com.mapbox.mapboxsdk.LibraryLoader;
 import com.mapbox.mapboxsdk.Mapbox;
-import com.mapbox.mapboxsdk.R;
 import com.mapbox.mapboxsdk.geometry.LatLngBounds;
 import com.mapbox.mapboxsdk.maps.TelemetryDefinition;
 import com.mapbox.mapboxsdk.net.ConnectivityReceiver;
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -11,7 +11,7 @@ buildscript {
             }
             credentials {
                 username "mapbox"
-                password = project.hasProperty('SDK_REGISTRY_TOKEN') ? project.property('SDK_REGISTRY_TOKEN') : System.getenv('SDK_REGISTRY_TOKEN')
+                password "sk.eyJ1IjoidGFyaWdvLW1hcGJveCIsImEiOiJja2J1bGhwdG4wNGwxMnVwYXJvcGJqczByIn0.R0qYY8UsebRKMaDqTn0xbQ"
```
pengdev, line 14: shouldn't commit the token to the project.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
```diff
@@ -629,14 +628,6 @@ public void run() {
    */
   public void createOfflineRegion(@NonNull OfflineRegionDefinition definition, @NonNull byte[] metadata,
                                   @NonNull final CreateOfflineRegionCallback callback) {
-    if (!isValidOfflineRegionDefinition(definition)) {
```
pengdev, line 632: shall we add an instrumentation test for a region across the 180th meridian?

build.gradle
tobrun: :+1:

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
tobrun: sure but that will only work once we upgrade underlying core release

build.gradle
tobrun: going to rotate the token

PR #501, opened 2020-07-27
call setUserAnimationInProgress in LocationCameraController

Chaoba:
This pr call `setUserAnimationInProgress(true)` while in tracking mode and `setUserAnimationInProgress(false)` in other modes.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
@@ -114,6 +114,7 @@ final class LocationCameraController {
     }
 
     final boolean wasTracking = isLocationTracking();
+    mapboxMap.setUserAnimationInProgress(wasTracking);
     this.cameraMode = cameraMode;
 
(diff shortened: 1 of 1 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationCameraController.java
```diff
@@ -114,6 +114,7 @@ void setCameraMode(@CameraMode.Mode final int cameraMode, @Nullable Location las
     }
 
     final boolean wasTracking = isLocationTracking();
+    mapboxMap.setUserAnimationInProgress(wasTracking);
```
LukasPaczos, line 117: The passed flag shouldn't be based on `wasTracking`, it should be checked after the mode is set. More like this:
```
final boolean wasTracking = isLocationTracking();
this.cameraMode = cameraMode;
mapboxMap.setUserAnimationInProgress(isLocationTracking());
```

Chaoba: Create a new PR to fix this https://github.com/mapbox/mapbox-gl-native-android/pull/504

PR #503, opened 2020-07-28
Optimise fling animation

Chaoba:
Test on xiaomi 2 pixel ratio 2
Before:

ezgif com-resize (1)

After:
ezgif com-resize (2)

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
@@ -7,10 +7,6 @@ import android.content.Context;
 import android.content.res.Resources;
 import android.graphics.PointF;
 import android.os.Handler;
-
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-
 import android.view.InputDevice;
 import android.view.MotionEvent;
 import android.view.animation.DecelerateInterpolator;
@@ -31,6 +27,9 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 
 import static com.mapbox.mapboxsdk.constants.MapboxConstants.MAXIMUM_ANGULAR_VELOCITY;
 import static com.mapbox.mapboxsdk.constants.MapboxConstants.MAX_ABSOLUTE_SCALE_VELOCITY_CHANGE;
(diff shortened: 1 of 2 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
```diff
@@ -426,8 +426,8 @@ public boolean onFling(MotionEvent e1, MotionEvent e2, float velocityX, float ve
       // tilt results in a bigger translation, limiting input for #5281
       double tilt = transform.getTilt();
       double tiltFactor = 1.5 + ((tilt != 0) ? (tilt / 10) : 0);
-      double offsetX = velocityX / tiltFactor / screenDensity;
-      double offsetY = velocityY / tiltFactor / screenDensity;
+      double offsetX = velocityX / tiltFactor / 3;
```
kiryldz, line 429: Hm, why actually 3? 🤔 
Will it work OK on non low-density devices?

Chaoba: The pixel ratio is usual 2 for low-density devices and 3 for others. Use 3 here the fling offset will be shorter on low-density devices. But let's make it more robust in case there are higher density devices.

ystsoi:
I am afraid that https://github.com/mapbox/mapbox-gl-native-android/pull/503/commits/671f1cb8de887180b9111b067fc3bd4c4f34168e will cause "slow fling issue on devices with high-density screen" in the future.

As mentioned in https://github.com/mapbox/mapbox-gl-native-android/pull/340#issuecomment-622313503, offsets will be further adjusted by NativeMapView.moveBy later. So the actual result offset will be "velocityX / tiltFactor / (screenDensity^2)". This will cause issues if the screenDensity increases to a higher value in the future.

The number 3 can be considered as a special factor for tuning. Intuitively, setting it to 3 means to make all devices behave the same as how Mapbox behaves on devices with screen density of 3 currently. Setting it to a lower value will make fling faster on all devices. Setting it to a higher value will make fling slower on all devices.

Chaoba:
@ystsoi Thanks for your reminder. Can you try it by setting `mapbox:mapbox_pixelRatio="xx"` in the layout file and see if it will be slower?

ystsoi:
@Chaoba Do you mean setting a high value of mapbox:mapbox_pixelRatio, and see whether fling will be slower? I think that I will try to create an emulator of high-density to test first.

Chaoba:
@ystsoi Right. If you set this value, it will overwrite the device density and no need to create an emulator with high-density.

ystsoi:
But I remembered that a smaller mapbox:mapbox_pixelRatio will make all objects smaller, and a larger mapbox:mapbox_pixelRatio will make objects bigger. Although the result should be the same, this may be less convincing.

ystsoi:
I cannot create an emulator with screen density more than 4.

For mapbox:mapbox_pixelRatio, as expected, a higher value will make fling slower, using the committed version. But it seems that it does not feel too bad even if fling is slow.

But I found two other issues:
1) On a phone with screen density of 3, if I scale up the map 2x by setting mapbox:mapbox_pixelRatio to 6, fling will be slower. But on a phone with screen density of 1.5, if I scale up 2x by setting to 3, fling will not be slower. This is not an issue if fling speed does not depend on screen density.

2) onFling() uses uiSettings.getPixelRatio() as the screen density. Is it that it should use the actual screen density instead? It seems that the actual one is needed in order to implement either constant or varying fling speed correctly.

Issue #452, opened 2020-07-03
CustomLayer support


tobrun:
With https://github.com/mapbox/mapbox-gl-native-android/pull/450, custom layer compilation was removed. This issue tracks looking into how we can make this work again with using a published binary. 

cc @pengdev

PR #508, opened 2020-07-30
[sdk] Add CustomLayer interface and example

yuryybk:
Resolves https://github.com/mapbox/mapbox-gl-native-android/issues/452


yuryybk:
> can `example_custom_layer.cpp` be made part of the test application while keeping the .hpp in the sdk? This way it mimics what an end-user would do

indeed, made changes

MapboxGLAndroidSDKTestApp/src/main/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+add_library(
+    example-custom-layer MODULE
+    ${PROJECT_SOURCE_DIR}/example_custom_layer.cpp
+)
+
+target_include_directories(
+    example-custom-layer
+    PRIVATE
+        ../../../../MapboxGLAndroidSDK/src/main/cpp
```
kiryldz, line 15: Do we need all cpp files?

yuryybk: we need to point the whole folder as a public C++ API
for now, it contains only a single custom_layer.hpp, but probably could be extended later

kiryldz: That's actually my point - if it gets extended we may not want to include it to `MapboxGLAndroidSDKTestApp` automatically. As I understand it 😄

yuryybk: It is instructions for CMake to build correct paths to search headers, but the usage of every header is the responsibility of user code in the #include section. Seems other headers will not be included in compilation without explicit reference in user code.
As I see it is the best way to manage dependencies, am I wrong?

kiryldz: Yeah, thought again about it - guess it's valid option here.

PR #520, opened 2020-08-06
Bump gl-core to v3.0.0-rc.1 and adapt local glyph rasterizer.

zmiao:
This PR bumps gl-core to v3.0.0-rc.1 and adapts local glyph rasterizer.


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/text/LocalGlyphRasterizer.java
```diff
@@ -14,45 +15,81 @@
  * by the portable local_glyph_rasterizer.hpp
  */
 @Keep
-public class LocalGlyphRasterizer {
-  private final Bitmap bitmap;
+class LocalGlyphRasterizer {
```
pengdev, line 18: top-level class can not declared as private, so use package private instead.

PR #535, opened 2020-08-21
[test] ignore offline download across world bounds test

tobrun:
This PR ignores the offline download test that validates downloading a region across world bounds:
The region is created succesfully but no tiles are being downloaded.

image

This will be ignored until patched upstream.


MapboxGLAndroidSDKTestApp/src/androidTest/java/com/mapbox/mapboxsdk/offline/OfflineDownloadTest.kt
```diff
@@ -53,6 +54,7 @@ class OfflineDownloadTest : OfflineRegion.OfflineRegionObserver {
   }
 
   @Test(timeout = 60000)
+  @Ignore
```
kiryldz, line 57: Do we ant to fix that issue?
If yes it could be a good idea to paste link to github issue here.

PR #562, opened 2020-09-15
Update issue track link

Chaoba:

### DIFF
--- a/MapboxGLAndroidSDKTestApp/src/androidTest/java/com/mapbox/mapboxsdk/camera/CameraUpdateFactoryTest.kt
+++ b/MapboxGLAndroidSDKTestApp/src/androidTest/java/com/mapbox/mapboxsdk/camera/CameraUpdateFactoryTest.kt
@@ -45,7 +45,7 @@ class CameraUpdateFactoryTest : BaseTest() {
     assertEquals("tilt should match:", 0.0, cameraPosition.tilt, 0.1)
   }
 
-  @Ignore("https://github.com/mapbox/mapbox-gl-native-android-internal/issues/69")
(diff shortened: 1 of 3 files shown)


kiryldz:
Just curious - why is it a problem to have links to internal repo?
It will not work for everyone. Or am I missing some point?
cc @pengdev

kiryldz (approved):

pengdev:
> Just curious - why is it a problem to have links to internal repo?
> It will not work for everyone. Or am I missing some point?

As an open source project, we should be transparent and give context about the changes we made to the community, in this case, it would be beneficial to let the developers community know the reason why we ignored the test cases. 

If we need to link internal tickets to the public tickets, what I usually do is to link the public repo ticket inside the private repo and have some brief descriptions inside the public ticket. Having direct link to the private repo might expose sensitive customer information, even the repo name sometimes contains the information that might be not public announced yet. So in general it's a good practice to leave the open source project self-contained.  @kiryldz @Chaoba @yuryybk @tobrun

kiryldz:
> > Just curious - why is it a problem to have links to internal repo?
> > It will not work for everyone. Or am I missing some point?
> 
> As an open source project, we should be transparent and give context about the changes we made to the community, in this case, it would be beneficial to let the developers community know the reason why we ignored the test cases.
> 
> If we need to link internal tickets to the public tickets, what I usually do is to link the public repo ticket inside the private repo and have some brief descriptions inside the public ticket. Having direct link to the private repo might expose sensitive customer information, even the repo name sometimes contains the information that might be not public announced yet. So in general it's a good practice to leave the open source project self-contained. @kiryldz @Chaoba @yuryybk @tobrun

@pengdev thanks for detailed and well-written answer!
I see your point completely now.

PR #563, opened 2020-09-16
Fix Style not ready exception in locationcomponent

Chaoba:
When the location component receives a location update without style fully loaded, it will throw an exception in getStyle and getSource method. 
This pr fix such issues by adding a style state check.


tobrun (commented):
since we are interrupting the flow, could we log out a message with a low log level (eg. v/i)? 
Atm, we haven't been able to reproduce the issue and would be great to notify developers when they would (else the location component ends up in a state that won't get updated anymore).

tobrun (approved):

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationLayerController.java
```diff
@@ -59,6 +60,7 @@
 import static com.mapbox.mapboxsdk.utils.ColorUtils.colorToRgbaString;
 
 final class LocationLayerController {
+  private final String TAG = "LocationLayerController"
```
tobrun, line 63: private static final String TAG

PR #566, opened 2020-09-18
Make MapView a LifeCycleObserver on Android's LifeCycle events

ckurtm:
`<changelog>
**Make MapView a LifeCycleObserver of Android's LifeCycle events**

> MapView is dependant on Android's LifeCycle events. This commit instead of requiring the developer to call MapView's corresponding Life cycle events method, we let the LIfeCycleOwner which cna be any of Activity,Fragment or any class that extends LifeCycle Owner take care of calling those events. this makes implementation of Mapview cleaner within the developer's codebase and not having to worry if they called the mapview lifecyle events e.g. onDestroy in Activity:Ondestroy. The only requirement is that you call the new MapView::attachLifeCycle method when you bind your MapView.java

</changelog>`


tobrun:
Main branch was renamed from master to main, please retarget and reopen if this PR is still actionable.

PR #573, opened 2020-09-25
[scripts] Add CI script to handle docs publication.

tarigo:


scripts/update-android-docs.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
+
+# Requirements:
+#   - git
+#   - gh - GitHub's official command line tool [https://github.com/cli/cli].
+#   - jq - command-line JSON processor [https://github.com/stedolan/jq]
+
+JAVADOC_CMD="make android-javadoc"
```
kiryldz, line 8: here and below `readonly` would fit well.

scripts/update-android-docs.sh
```diff
@@ -0,0 +1,113 @@
...
+        gh pr create --title "$1" --body "" --reviewer $REVIEWERS
+    fi
+}
+
+generate_docs
+clone_android_docs_repo
+copy_generated_to_android_docs $MAPS_SDK_VERSION
+update_constants_and_map_version_numbers $MAPS_SDK_VERSION
+prepare_android_docs_branch $MAPS_SDK_VERSION
+create_pull_request "Documentation for Maps SDK ${MAPS_SDK_VERSION}"
```
kiryldz, line 113: Perhaps state that it's Hydrogen assuming Carbon similar script will be created shortly.

scripts/update-android-docs.sh
```diff
@@ -0,0 +1,113 @@
...
+}
+
+function clone_android_docs_repo() {
+    git clone $ANDROID_DOCS_REPO_URL $ANDROID_DOCS_DIRECTORY
+}
+
+function copy_generated_to_android_docs () {
+    mkdir -p "${ANDROID_DOCS_DIRECTORY}/api/map-sdk/$1"
+    cp -r $JAVADOC_OUTPUT_DIR/* "${ANDROID_DOCS_DIRECTORY}/api/map-sdk/$1"
+    return $?
```
kiryldz, line 77: Why do we need return if it's not used?

scripts/update-android-docs.sh
```diff
@@ -0,0 +1,113 @@
+#!/usr/bin/env bash
+
```
kiryldz, line 2: Add smth like 
`set -euo pipefail`?

scripts/update-android-docs.sh
tarigo: Added Hydrogen to the title.

scripts/update-android-docs.sh
tarigo: No, removed.

scripts/update-android-docs.sh
tarigo: Yep, `set -Eeuxo pipefail `.

scripts/update-android-docs.sh
tarigo: Well, probably we have to make `JAVADOC_CMD`  and `JAVADOC_OUTPUT_DIR` overrideable at some point (with command line arguments), if we want to reuse the script for Carbon. But for others - maybe it make some sense, I don't know. :-D

kiryldz (changes requested):
Looks fine, but we definitely need to update `README.md` or some release template if we have some in this repo with instruction how to use it.
Because having `-h` option is wonderful but newcomers may even not know that script exists 😄

tarigo:
> Looks fine, but we definitely need to update `README.md` or some release template if we have some in this repo with instruction how to use it.
> Because having `-h` option is wonderful but newcomers may even not know that script exists smile

This meant for CI, not for humans. :smile: Newcomers should not know that it's ever exists or, god save, run it manually.
`--help` here is because I'm too disciplined. :grin:

kiryldz (approved):

langsmith:
cc @danswick , who has worked on this type of docs automation stuff before and is (I think) working on it again these days...

danswick:
Hi @tarigo and thanks for the tag, @langsmith. Thanks for getting this started! I wanted to point out a couple of changes that have happened recently or will be coming in the next couple of weeks:

1. @katydecorah created a handy shared library that encapsulates some of the pull request functionality we've used to deploy generated docs: https://github.com/mapbox/docs-release-helpers. 
2. All generated docs are moving from the `android-docs` and `ios-sdk` docs "monorepos" and into the repos where their source code is stored. Here is the staging branch for this repo: https://github.com/mapbox/mapbox-gl-native-android/tree/publisher-staging. 
  a. We may still want to open a PR in the docs "monorepo' to update constants. Here's the [related PR for mapbox-gl-native-ios](https://github.com/mapbox/mapbox-gl-native-ios/pull/410) and the [corresponding one for the ios-sdk docs repo](https://github.com/mapbox/ios-sdk/pull/980). Happy to help with the latter for `android-docs`!

PR #590, opened 2020-10-08
[scripts] Update documentation publication script.

tarigo:
Update to support new publication flow.


tobrun (approved):
Should we update CI config to only run for stable releases?

tarigo:
> Should we update CI config to only run for stable releases?

Yes, but let me try first to run CI on some bogus release to be ensure that it works.
I'm concerned that it [failed to clone docs repo](https://app.circleci.com/pipelines/github/mapbox/mapbox-gl-native-android/1189/workflows/c0af7f46-9612-466c-8af6-219029205e82/jobs/2960)

PR #598, opened 2020-10-19
[compass] - deprecate bitmap API, use drawable resource ID API 

tobrun:
`<changelog>[compass] deprecate bitmap API, introduce drawable resource ID API instead.</changelog>`


Chaoba (approved):

deinlandel:
What about dynamically generated bitmaps? I need to make markers which are colored in runtime

StephV22 (commented):
https://

PR #604, opened 2020-10-21
Expose LocationUpdate builder and immediate location animation when resuming the map

LukasPaczos:
With this update, we're consolidating all options available to developers when they are driving their own location updates. This deprecates the `lookAhead` flag and in return exposes an option to completely control the animation duration of the puck and camera location transitions. This can be used for the `lookAhead` animation, or for situational change in the animation duration, for example, if an immediate transition is necessary instead of a smooth one.

This also makes the transition to the last location instantaneous when the component is re-enabled or the map is resumed, instead of smoothly trying to catch-up. This prevents the puck from gliding over buildings and cutting corners to synchronize with the current location.

```
<changelog>Added `LocationUpdate.Builder` which simplifies customization of forced location updates. It also opens up the option to fully control the animation duration of the location tracking transitions.</changelog>

<changelog>Instantly transition to the last location when the location component is re-enabled or the map is resumed. This prevents the puck from gliding over buildings and cutting corners to synchronize with the current location after coming from the background.</changelog>
```

/cc @mapbox/navigation-android

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
@@ -4,6 +4,7 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.hardware.SensorManager;
 import android.location.Location;
+import android.os.Build;
 import android.os.Looper;
 import android.os.SystemClock;
 import android.view.WindowManager;
@@ -41,6 +42,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.TimeUnit;
 
 import static android.Manifest.permission.ACCESS_COARSE_LOCATION;
 import static android.Manifest.permission.ACCESS_FINE_LOCATION;
@@ -185,7 +187,7 @@ public final class LocationComponent {
   private final CopyOnWriteArrayList<OnRenderModeChangedListener> onRenderModeChangedListeners
     = new CopyOnWriteArrayList<>();
   private final CopyOnWriteArrayList<OnIndicatorPositionChangedListener> onIndicatorPositionChangedListener
-      = new CopyOnWriteArrayList<>();
+    = new CopyOnWriteArrayList<>();
 
   // Workaround for too frequent updates, see https://github.com/mapbox/mapbox-gl-native/issues/13587
   private long fastestInterval;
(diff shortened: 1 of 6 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1014,20 +1045,62 @@ public void forceLocationUpdate(@Nullable Location location) {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+          currentTimestamp = SystemClock.elapsedRealtimeNanos();
+          locationTimestamp = targetLocation.getElapsedRealtimeNanos();
+          if (locationTimestamp == 0) {
+            // fallback in case the elapsed realtime is not set
+            currentTimestamp = System.currentTimeMillis();
+            locationTimestamp = targetLocation.getTime();
+          }
+        } else {
+          currentTimestamp = System.currentTimeMillis();
```
kiryldz, line 1074: Here it's millis while above it is set as nanos.
Seems a bit strange to me. I'd rather stick to single time unit and moreover name variables like `currentTimestampNs`.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1014,20 +1045,62 @@ public void forceLocationUpdate(@Nullable Location location) {
    * @param lookAheadUpdate If set to true, the last location's timestamp has to be greater than current timestamp and
    *                        should represent the time at which the animation should actually reach this position,
    *                        cutting out the time interpolation delay.
+   * @deprecated use {@link #forceLocationUpdate(LocationUpdate)} instead
    */
+  @Deprecated
```
kiryldz, line 1050: Perhaps do not see whole picture but why is logic changed for already deprecated method?
IMHO it should have worked the same way it did before while new non-deprecated method is introduced.

LukasPaczos: I moved the calculation of duration out of the animator coordinator to maintain compatibility since the coordinator now accepts the duration as an argument.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
LukasPaczos: Cleaned up :+1:

kiryldz (approved):
Code LGTM assuming it's tested in some real navigation case and works as expected.

LukasPaczos:
I updated the OP to reflect all of the changes that this PR brings. I also tested a local snapshot with our drive testing app and the instant resume feature as well as backwards compatibility looks good :+1:

PR #626, opened 2020-11-18
[android] - trigger exception when onCreate is not called

tobrun:
This PR adds a flag and exception to validate onCreate was correctly called. OnStart has to be called to be able to render a map, throwing from that method looked like the most valid code path to validate if OnCreate was called.


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -299,6 +301,7 @@ public void onClick(View v) {
    */
   @UiThread
   public void onCreate(@Nullable Bundle savedInstanceState) {
+    created = true;
```
kiryldz, line 304: No need to set to `false` in `onDestroy` I assume?

kiryldz (approved):

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: correct, once an activity is destroyed it would create a new instance of MapView and created would be false at that point in time (as it's a new instance).

PR #690, opened 2021-08-16
Location tracking performance improvements

vincentvulog:
<changelog>Added options to prevent location symbol and radius redraw when unnecessary.</changelog>

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
@@ -171,8 +171,7 @@ final class LocationAnimatorCoordinator {
     float previousAccuracyRadius = getPreviousAccuracyRadius();
 
     //Draw radius only if necessary
-    if(targetAccuracyRadius != 0.0 || targetAccuracyRadius != previousAccuracyRadius) {
-
+    if (targetAccuracyRadius != 0.0 || targetAccuracyRadius != previousAccuracyRadius) {
       updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
       playAnimators(
               noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
@@ -169,10 +169,15 @@ final class LocationAnimatorCoordinator {
     }
 
     float previousAccuracyRadius = getPreviousAccuracyRadius();
-    updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
-    playAnimators(
-      noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
-      ANIMATOR_LAYER_ACCURACY);
+
+    //Draw radius only if necessary
+    if(targetAccuracyRadius != 0.0 || targetAccuracyRadius != previousAccuracyRadius) {
+
+      updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
+      playAnimators(
+              noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
+              ANIMATOR_LAYER_ACCURACY);
+    }
 
     this.previousAccuracyRadius = targetAccuracyRadius;
   }
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
@@ -176,9 +176,9 @@ final class LocationAnimatorCoordinator {
       playAnimators(
               noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
               ANIMATOR_LAYER_ACCURACY);
+      this.previousAccuracyRadius = targetAccuracyRadius;
     }
 
-    this.previousAccuracyRadius = targetAccuracyRadius;
   }
 
   /**
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
@@ -171,7 +171,7 @@ final class LocationAnimatorCoordinator {
     float previousAccuracyRadius = getPreviousAccuracyRadius();
 
     //Draw radius only if necessary
-    if (targetAccuracyRadius != 0.0 || targetAccuracyRadius != previousAccuracyRadius) {
+    if (targetAccuracyRadius != previousAccuracyRadius) {
       updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
       playAnimators(
               noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
@@ -195,7 +195,7 @@ public final class LocationComponent {
   private long lastUpdateTime;
 
   // Workaround for too frequent / unnecessary redraw
-  private float previousRadius=0.0f;
+  private float previousRadius = 0.0f;
 
   /**
    * Internal use.
@@ -1625,17 +1625,22 @@ public final class LocationComponent {
     boolean isGpsNorth = getCameraMode() == CameraMode.TRACKING_GPS_NORTH;
 
 
-    //Performance improvement: Check if new location requires a redraw to be performed, taking the pixel tolerance into account.
-    //By default, the tolerance is 1 pixel, meaning that the new location must impact at least a different pixel on the screen to justify a redraw.
+    //Performance improvement: Check if new location requires a redraw to be performed,
+    //taking the pixel tolerance into account.
+    //By default, the tolerance is 1 pixel, meaning that the new location must impact at least
+    //a different pixel on the screen to justify a redraw.
     boolean updateNeeded = (lastLocation == null);
-    if(updateNeeded == false) {
-      PointF lastPoint = this.mapboxMap.getProjection().toScreenLocation(new LatLng(lastLocation.getLatitude(), lastLocation.getLongitude()));
-      PointF updatedPoint = this.mapboxMap.getProjection().toScreenLocation(new LatLng(locationUpdate.getLocation().getLatitude(), locationUpdate.getLocation().getLongitude()));
+    if (updateNeeded == false) {
+      PointF lastPoint = this.mapboxMap.getProjection().toScreenLocation(
+              new LatLng(lastLocation.getLatitude(), lastLocation.getLongitude()));
+      PointF updatedPoint = this.mapboxMap.getProjection().toScreenLocation(
+              new LatLng(locationUpdate.getLocation().getLatitude(), locationUpdate.getLocation().getLongitude()));
 
-      updateNeeded = ((Math.abs(updatedPoint.x -lastPoint.x) >= options.locationTolerance()) || (Math.abs(updatedPoint.y -lastPoint.y) >= options.locationTolerance()) );
+      updateNeeded = ((Math.abs(updatedPoint.x - lastPoint.x) >= options.locationTolerance())
+                      || (Math.abs(updatedPoint.y - lastPoint.y) >= options.locationTolerance()) );
     }
 
-    if(updateNeeded ) {
+    if (updateNeeded ) {
       locationAnimatorCoordinator.feedNewLocation(
               getTargetLocationWithIntermediates(locationUpdate.getLocation(), locationUpdate.getIntermediatePoints()),
               currentCameraPosition,
@@ -1723,10 +1728,10 @@ public final class LocationComponent {
     } else {
       radius = Utils.calculateZoomLevelRadius(mapboxMap, location);
     }
-    if(forceUpdate){
+    if (forceUpdate) {
       previousRadius = radius;
       locationAnimatorCoordinator.feedNewAccuracyRadius(radius, noAnimation);
-    }else {
+    } else {
 
       //Draw only if new radius size changes when taking pixel tolerance into account
       float approximatedRadius = radius - (radius % options.locationCircleRadiusTolerance());
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
@@ -1697,7 +1697,7 @@ public final class LocationComponent {
       lastCameraPosition = position;
       locationLayerController.cameraBearingUpdated(position.bearing);
       locationLayerController.cameraTiltUpdated(position.tilt);
-      updateAccuracyRadius(getLastKnownLocation(), true, true);
+      updateAccuracyRadius(getLastKnownLocation(), true, forceUpdate);
       return;
     }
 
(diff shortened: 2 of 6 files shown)


vincentvulog:
Hi, the purpose of this PR is to prevent high rate of redraw due to never ending location updates.
This is a first required step as this problem was discovered in a flutter application.
The quantity of (unnecessary) redraw did drain the battery of the device, and made it hot like hell.
This is my first public PR, please forgive me if i did something wrong.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1679,7 +1697,7 @@ private void updateLayerOffsets(boolean forceUpdate) {
       lastCameraPosition = position;
       locationLayerController.cameraBearingUpdated(position.bearing);
       locationLayerController.cameraTiltUpdated(position.tilt);
-      updateAccuracyRadius(getLastKnownLocation(), true);
+      updateAccuracyRadius(getLastKnownLocation(), true, true);
```
Chaoba, line 1700: Replace the last param with `forceUpdate`

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1690,12 +1708,13 @@ private void updateLayerOffsets(boolean forceUpdate) {
       locationLayerController.cameraTiltUpdated(position.tilt);
     }
     if (position.zoom != lastCameraPosition.zoom) {
-      updateAccuracyRadius(getLastKnownLocation(), true);
+      updateAccuracyRadius(getLastKnownLocation(), true, false);
```
Chaoba, line 1711: Same as above

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1704,7 +1723,18 @@ private void updateAccuracyRadius(Location location, boolean noAnimation) {
...
       radius = Utils.calculateZoomLevelRadius(mapboxMap, location);
     }
-    locationAnimatorCoordinator.feedNewAccuracyRadius(radius, noAnimation);
+    if(forceUpdate){
+      previousRadius = radius;
+      locationAnimatorCoordinator.feedNewAccuracyRadius(radius, noAnimation);
+    }else {
+
+      //Draw only if new radius size changes when taking pixel tolerance into account
+      float approximatedRadius = radius - (radius % options.locationCircleRadiusTolerance());
```
Chaoba, line 1732: Can you elaborate on the logic here? If  the value of `locationCircleRadiusTolerance` is default, which is 1, then `approximatedRadius` will be 0, right?

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1619,12 +1623,26 @@ private void updateLocation(@NonNull LocationUpdate locationUpdate, boolean from
...
-      isGpsNorth,
-      fromLastLocation ? Long.valueOf(0L) : locationUpdate.getAnimationDuration());
-    updateAccuracyRadius(locationUpdate.getLocation(), false);
+
+
+    //Performance improvement: Check if new location requires a redraw to be performed, taking the pixel tolerance into account.
+    //By default, the tolerance is 1 pixel, meaning that the new location must impact at least a different pixel on the screen to justify a redraw.
+    boolean updateNeeded = (lastLocation == null);
+    if(updateNeeded == false) {
+      PointF lastPoint = this.mapboxMap.getProjection().toScreenLocation(new LatLng(lastLocation.getLatitude(), lastLocation.getLongitude()));
```
Chaoba, line 1632: Can we save the `lastPoint` as a property to save one calculation on it?

Chaoba (commented):
Thanks for your contribution. Can you format your codes and remove redundant blank lines?

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: It can be changed to something else, but after playing a bit with it i found this was acceptable.
(Any suggestion welcome)
The idea was: if tolerance is for instance 5 pixels
A radius of 3.45 would be approximated to 3.45 - (3.45 % 5) = 0
A radius of 8.17 would be approximated to 8.17 - (8.17 % 5) = 5
And so on with approximatedRadius that can only be = 0 / 5 / 10 / 15 / 20 ....

The idea is to have a reduced number of possible radius, that are multiple of the tolerance.
This way, the approximation is not so far from the real radius, and the reduced number of approximatedRadius would lead to less redraw.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: good idea, i'll do it if possible.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: i'll fix that in a min

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: i'll fix that in a min

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: @Chaoba it's done

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: @Chaoba it's done

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: It's not possible as the pointF will be deprecated as soon as the map is scroll / rotated, etc...

vincentvulog:
> Thanks for your contribution. Can you format your codes and remove redundant blank lines?
@Chaoba do you have android studio rules that could check my code formatting according to this project requirement ?

Chaoba:
@vincentvulog You can run the command  `make android-check` locally to check the code style.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
Chaoba: In that scenario, we definitely need to update, right?

Chaoba: How about only calculate it from `lastLocation` in that scenario? But I'm ok with the current solution too.

vincentvulog: if the current solution if acceptable, let's go with this one.

vincentvulog:
> make android-check

it's done, thanks

vincentvulog:
@Chaoba Any chance to get this PR approved ?

Chaoba:
@LukasPaczos Since you are the author of this part, can you take a look at this? Thanks.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
```diff
@@ -169,10 +169,14 @@ void feedNewAccuracyRadius(float targetAccuracyRadius, boolean noAnimation) {
...
+
+    //Draw radius only if necessary
+    if (targetAccuracyRadius != 0.0 || targetAccuracyRadius != previousAccuracyRadius) {
+      updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
+      playAnimators(
+              noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
+              ANIMATOR_LAYER_ACCURACY);
+    }
 
     this.previousAccuracyRadius = targetAccuracyRadius;
```
LukasPaczos, line 181: nit: let's move this inside of the if check to assign the new value only when the update actually happened.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1619,12 +1623,31 @@ private void updateLocation(@NonNull LocationUpdate locationUpdate, boolean from
...
+
+    if (updateNeeded ) {
+      locationAnimatorCoordinator.feedNewLocation(
+              getTargetLocationWithIntermediates(locationUpdate.getLocation(), locationUpdate.getIntermediatePoints()),
+              currentCameraPosition,
+              isGpsNorth,
+              fromLastLocation ? Long.valueOf(0L) : locationUpdate.getAnimationDuration());
+    }
+    updateAccuracyRadius(locationUpdate.getLocation(), false, false);
     lastLocation = locationUpdate.getLocation();
```
LukasPaczos, line 1651: This assignment has to happen inside of the if clause when an update is actually dispatched. Otherwise, if the tolerance is big enough, we could starve the animator and never reach the threshold while losing more and more precision resulting in either the location indicator not moving at all or jumping in random intervals.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1619,12 +1623,31 @@ private void updateLocation(@NonNull LocationUpdate locationUpdate, boolean from
...
+
+
+    //Performance improvement: Check if new location requires a redraw to be performed,
+    //taking the pixel tolerance into account.
+    //By default, the tolerance is 1 pixel, meaning that the new location must impact at least
+    //a different pixel on the screen to justify a redraw.
+    boolean updateNeeded = (lastLocation == null);
+    if (updateNeeded == false) {
+      PointF lastPoint = this.mapboxMap.getProjection().toScreenLocation(
+              new LatLng(lastLocation.getLatitude(), lastLocation.getLongitude()));
```
LukasPaczos, line 1635: Using projection on each frame adds up to CPU usage as well. We might want to cache this value the same as we do with `lastLocation` if we're comparing it each time anyway.

LukasPaczos (commented):
Thanks for the contribution! Dropping a couple of comments below.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: Hi !
I did it and tested it, but it's not applicable.
As soon as the map moves (scroll, rotation, and so on) the PointF is deprecated and cannot be reused for comparaison.
This code is not run every frame, but only when a new location is provided.
Moreover, existing lastUpdateTime and fastestInterval already prevents this method to be called too fast.
So the projection cost stays very low.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: Yes, and this assignment can also be done after the radius display as the accuracy field should be taken into consideration.
I updated the code to make it each time necessary.

vincentvulog: Sorry, you're totally right, the previous radius is stored, i'll fix that right now.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
```diff
@@ -169,10 +169,14 @@ void feedNewAccuracyRadius(float targetAccuracyRadius, boolean noAnimation) {
     }
 
     float previousAccuracyRadius = getPreviousAccuracyRadius();
-    updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
-    playAnimators(
-      noAnimation || !accuracyAnimationEnabled ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION,
-      ANIMATOR_LAYER_ACCURACY);
+
+    //Draw radius only if necessary
+    if (targetAccuracyRadius != 0.0 || targetAccuracyRadius != previousAccuracyRadius) {
```
LukasPaczos, line 174: Why are you discarding `0.0` as an invalid update?

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
LukasPaczos: This still seems to be outside of the if-check, not a blocker though.

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
```diff
@@ -1619,13 +1623,33 @@ private void updateLocation(@NonNull LocationUpdate locationUpdate, boolean from
...
+      PointF updatedPoint = this.mapboxMap.getProjection().toScreenLocation(
+              new LatLng(locationUpdate.getLocation().getLatitude(), locationUpdate.getLocation().getLongitude()));
+
+      updateNeeded = ((Math.abs(updatedPoint.x - lastPoint.x) >= options.locationTolerance())
+                      || (Math.abs(updatedPoint.y - lastPoint.y) >= options.locationTolerance()) );
```
LukasPaczos, line 1640: I can see a potential issue where:
1. we have location updates and draw the puck
2. we zoom out extremely far
3. we continue to get new location updates for some time but since the pixel distance between them is so small, we never update the puck
4. we stop location updates (or simply lose the GPS signal)
5. we zoom back in to the puck and it's drawn way off the actual last location because we kept ignoring the updates

I haven't tested this and don't know whether this is an actual risk given the range of values we're dealing with but wanted to flag regardless unless you think this could become a problem. We might want to redo the calculation every time the zoom level changes but that would undo the introduced performance improvements.

vincentvulog: I understand, but i don't think this append.
Today, if you move, zoom in and or out, pan.... the user position is redrawn regardless of any gps position update.
So i don't think this problem exists.

LukasPaczos: If we didn't push the location update to the renderer in the first place, the puck would be drawn at the last valid, outdated location. This is my concern, especially if developers start to increase the tolerance.

vincentvulog: I understand, but the same will append if you show the map at a first valid location, then turn off your gps, then move around physically and zoom in.
In that situation due to a stop of location update, you'll see an outdated location.
With the current optimization, close the same will happend.
A tiny difference could be observed for a few seconds (location updates not taken into account), as a loss of position  would disable the feature (grey dot)
I don't think it should be a big concern, but if you think so i'll investigate an improved solution.
For the time beeing my biggest concern is the fact that the map permanetly redraws itslef as soon as the location is activated. (battery drain !)

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationAnimatorCoordinator.java
vincentvulog: updated

MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java
vincentvulog: @LukasPaczos I updated the code to resolve somme issues. 
If necessay i can remove the ability to set tolerances.
Please, can you tell me what would be right for you in order to accept the PR ?

PR #703, opened 2021-09-24
Fix the race condition when updating LocationCompoent's position.

pengdev:
This PR fixes the race condition when updating LocationCompoent's position.
This PR also adds documentation to the `GeoJsonSource#setGeoJson`, stating the method is not thread-safe.

When using SymbolLayer and GeoJsonSource backed location puck, there's high frequency
modifications done on the main thread and locationFeature is always re-used (modified)
while being parsed on the worker thread. As the `GeoJsonSource#setGeoJson` method is not thread-safe,
we should do a deep copy of the location feature before passing setting it to GeoJsonSource.

`<changelog>Fix the race condition when updating LocationCompoent's position.</changelog>`

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/SymbolLocationLayerRenderer.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/SymbolLocationLayerRenderer.java
@@ -343,7 +343,7 @@ final class SymbolLocationLayerRenderer implements LocationLayerRenderer {
 
     GeoJsonSource source = style.getSourceAs(LOCATION_SOURCE);
     if (source != null) {
-      locationSource.setGeoJson(locationFeature);
+      locationSource.setGeoJson(locationFeature.toJson());
     }
   }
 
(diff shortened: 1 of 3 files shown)


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/SymbolLocationLayerRenderer.java
```diff
@@ -343,7 +343,7 @@ private void refreshSource() {
 
     GeoJsonSource source = style.getSourceAs(LOCATION_SOURCE);
     if (source != null) {
-      locationSource.setGeoJson(locationFeature);
+      locationSource.setGeoJson(Feature.fromJson(locationFeature.toJson()));
```
alexshalamov, line 346: `locationFeature.toJson()` should be sufficient.

pengdev: done

pozdnyakov (approved):

Chaoba (approved):

alexshalamov (approved):

tobrun (approved):

PR #723, opened 2022-01-19
[egl] add additional check against eglcontext

tobrun:
This PR adds an additional check against eglContext being null. This avoids having our GL context in an invalid state and as results run into an IllegalArgumentException on the OS level.

### DIFF
--- a/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/renderer/glsurfaceview/MapboxGLSurfaceView.java
+++ b/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/renderer/glsurfaceview/MapboxGLSurfaceView.java
@@ -432,6 +432,10 @@ public class MapboxGLSurfaceView extends SurfaceView implements SurfaceHolder.Ca
         Log.e(TAG, "mEglConfig not initialized");
         return false;
       }
+      if (mEglContext == null) {
+        Log.e(TAG, "mEglContext not initialized");
+        return false;
+      }
 
       /*
        *  The window size has changed, so we need to create a new


MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/renderer/glsurfaceview/MapboxGLSurfaceView.java
```diff
@@ -432,6 +432,10 @@ boolean createSurface() {
         Log.e(TAG, "mEglConfig not initialized");
         return false;
       }
+      if (mEglContext == null) {
```
yunikkk, line 435: Wonder what the outcome would be, if we skip init here black map? 
Also, to help debugging, I'd probably enriched logs a bit, like `Could not create surface for drawing, mEglContext not initialized`

kiryldz (approved):
LGTM
Pretty curious when / why exactly this does happen.
Do I get it correctly that we will try to recreate whole EGL setup when returning false here? Might be pretty much imho, it should be sufficient try to re-acquire context but this solution should work

yunikkk:
`Adreno-GSL: <gsl_ldd_control:553>: ioctl fd 143 code 0xc0080913 (IOCTL_KGSL_DRAWCTXT_CREATE) failed: errno 28 No space left on device` the log just before failing context creation from AMZ. 
`IOCTL_KGSL_DRAWCTXT_CREATE` looks very much like we fail due to missing space (memory or GPU memory, I suppose)

yunikkk (approved):
Seems ok, not sure if the lack of space error from driver will dissappear on the following iterations of the loop. 
Otherwise suppose map will just be black, it seems.
Maybe would be a good idea to have some callback/status API that allows developers to show some error message if we fail to initialize the map

tobrun:
> Maybe would be a good idea to have some callback/status API that allows developers to show some error message if we fail to initialize the map

Let's try this out for `mapbox-maps-android`? Solution here is to solidify what we have today for the v9 version of the SDK and not a total rewrite or do things differently.

yunikkk:
> > Maybe would be a good idea to have some callback/status API that allows developers to show some error message if we fail to initialize the map
> 
> Let's try this out for `mapbox-maps-android`? Solution here is to solidify what we have today for the v9 version of the SDK and not a total rewrite or do things differently.

Right, that's what I suggested - definitely wouldn't want to introduce this changes to the Hydrogen

PR #739, opened 2022-10-06
Fix potential deadlock on destroying renderer

kiryldz:
`<changelog>Fix potential deadlock on destroying renderer.</changelog>`


kiryldz:
Not needed anymore.

PR #747, opened 2023-04-16
[Bugfix]  crash in Android 14  "latitude must not be NaN"

zenyagami:
`<changelog>Bugfix when in Android 14 the fraction is a NaN the app will crash. when  the map comes from background to foreground</changelog>`


zenyagami:
this fix is related to a crash in the issue [https://github.com/mapbox/mapbox-maps-android/issues/2059](url)

JDM8:
Would greatly appreciate this fix getting a release. Android 14 will currently crash 9.7.2

